PR #273, opened 2013-06-07
Recommendations

monken:
@oalders, @yanick 
Let's discuss here


oalders:
Closing for now.  We can re-open if there's future activity on this.

PR #312, opened 2014-05-01
Mo/es1.0.0

monken:
**Do not merge yet**
I want to keep this open as a todo list for Elasticsearch 1.x.x compatibility

- [ ] make all tests pass
- [ ] explain breaking changes and /v1/ interface
- [ ] make `fields` query param mandatory for future backwards compat and warnings
- [ ] discuss changes in test files, ideally there shouldn't be any change
- [ ] set up 2nd metacpan server fully indexed to allow testing of metacpan.org and other clients


ranguard:
Closing as we have started this in other branches

PR #375, opened 2015-01-08
Check for undefined symlinks in a release and skip them

andreeap:


oalders:
The fix looks good, but it feels like the kind of fix that could use a regression test.  You could add an archive to the fake CPAN in the test suite which demonstrates this particular problem.  Then see if this archive is indexed successfully.

Alternatively, you could write a test that uses https://metacpan.org/pod/Test::RequiresInternet  This would download the tarball that we're having problems with and do the recursive search through the tarball without errors.  Doing this would mean that you'd need to break out the File::Find logic into a subroutine of its own or even into a different module.  That would be a good thing, though, as it would take some of the complexity out of the current subroutine, which is too big.

I'm reluctant to add actual distributions to the test suite if we can avoid it, but using the above approach would mean that the test runs if there's access to the Internet and is otherwise skipped, which seems like a good enough approach.

andreeap:
I already decided to break out the subroutine as we talked on Monday, so I will go with the second option.

oalders:
@andreeap why did you close this pull request?

andreeap:
@oalders I think I did this by mistake, I'll reopen it when I can push everything I worked on

oalders:
@andreeap I think it can only be opened once there are new commits.

andreeap:
I had new commits on master and still couldn't open it.

PR #378, opened 2015-01-14
add new module ProcessTarball instead of import_tarball

andreeap:
@oalders I can't reopen the last pull request, I did a new one to see if tests are failing for what I have changed

### DIFF
--- a/docs/API-docs.md
+++ b/docs/API-docs.md
@@ -8,7 +8,7 @@ _All of these URLs can be tested using the [MetaCPAN Explorer](http://explorer.m
 
 To learn more about the ElasticSearch query DSL check out Clinton Gormley's [Terms of Endearment - ES Query DSL Explained] (http://www.slideshare.net/clintongormley/terms-of-endearment-the-elasticsearch-query-dsl-explained) slides.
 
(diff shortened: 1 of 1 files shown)


oalders:
@andreeap where is the test file you added?

andreeap:
I have not included that yet, it is included on my forked repository,
because the way I changed the files should still be passing the tests, but
I think I was wrong.

oalders:
If you don't include it, we can't help you with it.

andreeap:
I added the test file on this branch too, but like I said on hackpad, I am not sure where to include this test. The logic in the test is  that I added an archive to fakecpan, and then verified how many files got indexed and compared to how many should have been.

oalders:
What do you mean by "where to include"?

andreeap:
I added this to t//script/release and I am not sure if it's ok. Do I have to do something else to be sure it will get into the test?

oalders:
When Travis runs, it will run anything under t/ that has a ".t" suffix, so this is fine.  As far as naming the test, we can look at that later.  The main thing is that it can be run.

andreeap:
I made a mistake when I named it, I can push the change, but I will let travis do its job first.

andreeap:
I have not closed this, is there something that closes pull requests automatically?

Issue #110, opened 2011-06-30
Makefiles are being indexed


oalders:
In my list of latest modules, some makefiles are appearing.  Check this out:

http://beta.metacpan.org/search?q=makefile.pl

oalders:
This looks to be fixed now.  :)

Issue #323, opened 2014-07-18
Clarify value of "authorized" in relation to "indexed"


rwstauner:
I fixed the display issue in https://github.com/CPAN-API/metacpan-web/issues/1251
but I still wonder if the api should be changed at all (or docs be clarified or something):

> However, if we were going to change the indexer/api data, i'm not sure what would be appropriate...
> if indexed is false, authorized could be true (b/c it's not _un_authorized) or maybe authorized should technically be null (undefined), though I'm not sure how ES feels about that.

If indexed is false, should consumers simply ignore authorized (we can leave it as is)?
Or to demonstrate that it should be ignored should we change the value?

rwstauner:
I had trouble finding this other comment (#373) so I'm quoting it here to collect the comments in one place:

> The documentation for "indexed"
> says "include in the search index", however that sounds different to me than the more verbose documentation for "set_indexed".
> It was once described to me that "indexed" essentially meant "not no_indexed by the META file"
> and that seems like roughly what "set_indexed" is doing.
> 
> The "authorized" value defaults to true and gets set to false if the module is in 06perms but the releaser is not associated with it.
> 
> So by those definitions both of these "cpanm" files from each dist would get marked true/true.
> That's obviously not very helpful, though, because in this case,
> while neither is explicitly not indexed (by META), nor not authorized (by 06perms),
> they are also not actually indexed (in 02packages) nor actually authorized (in 06perms)...
> So I'm still not sure what the right answer is.
> If the pod files were indexed:false and authorized:false they wouldn't be searchable on the web site.
> 
> I'd love to come up with a clearer definition for these values.
> If we can come up with better attributes we could put it in the next version of the API.

oalders:
After discussing this today, I really do think we need to come up with something better.  It was confusing for me to explain and after re-reading this I realize that it's actually _just confusing_.

ranguard:
We want to work out WHAT we should do, even if we don't actually do it at MetaHack

tsibley:
I've pushed some docs on [the state of these very confusing fields][doc] to the branch [tsibley/document-indexer-flags](https://github.com/metacpan/metacpan-api/tree/tsibley/document-indexer-flags).

I _haven't_ carefully audited queries for the ways these fields are used the search, `download_url` endpoint, etc, but the fields are used in many queries.

[...]

Issue #100, opened 2011-06-03
Author search returns no results


oalders:
This search works when run without the proxy (tested on my version of the API from last week).  Returns nothing when run against the current proxy.

```
curl -XGET 'http://api.beta.metacpan.org:80/v0/author/_search?scroll=5m&pretty=1'  -d '
{
   "query" : {
      "match_all" : {}
   },
   "size" : 100
}
'
```

PR #385, opened 2015-02-01
Files listed under other files are not authorized

andreeap:
This is a new pull request for #110 because I cleaned my commit history and rebased and somehow the pull request got closed automatically. This commit is ready for review.

### DIFF
--- a/t/document/file.t
+++ b/t/document/file.t
@@ -217,6 +217,26 @@ subtest 'Packages starting with underscore are not indexed' => sub {
     is( $file->module->[0]->indexed, 0, 'Package is not indexed' );
 };
 
+subtest 'files listed under other files' => sub {
+    my $content = <<'END';
+=head1 NAME
+
+Makefile.PL - configure Makefile
+
+=head1 DESCRIPTION
+
+just a makefile description
+
+END
+    my $file = new_file_doc(
+        name       => 'Makefile.PL',
+        content_cb => sub { \$content }
+    );
+
+    $file->set_indexed( {} );
+    is( $file->indexed, 0, 'File listed under other files is not indexed' );
+};
+
 subtest 'pod name/package mismatch' => sub {
     my $content = <<'END';
 package
--- a/t/document/file.t
+++ b/t/document/file.t
@@ -233,7 +233,6 @@ END
         content_cb => sub { \$content }
     );
 
-    $file->set_indexed( {} );
     is( $file->indexed, 0, 'File listed under other files is not indexed' );
 };
 
--- a/cpanfile
+++ b/cpanfile
@@ -77,6 +77,7 @@ requires 'JSON', '2.90';
 requires 'LWP::Protocol::https';
 requires 'LWP::UserAgent';
 requires 'LWP::UserAgent::Paranoid';
+requires 'List::AllUtils';
 requires 'List::MoreUtils';
 requires 'List::Util';
 requires 'Log::Contextual';
--- a/cpanfile.snapshot
+++ b/cpanfile.snapshot
@@ -4020,6 +4020,18 @@ DISTRIBUTIONS
     requirements:
       Test::More 0
       version 0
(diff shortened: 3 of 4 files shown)


oalders:
To me, this looks to be ready to merge.  @rwstauner any comments?

t/document/file.t
```diff
@@ -212,6 +212,13 @@ subtest 'Packages starting with underscore are not indexed' => sub {
     is( $file->module->[0]->indexed, 0, 'Package is not indexed' );
 };
 
+subtest 'files listed under other files' => sub {
+    my $file = new_file_doc( name => 'Makefile.PL' );
+    $file->set_authorized( $file->module => $file->author );
```
rwstauner: These arguments `( $file->module => $file->author )` are confusing as they don't match the sub definition.
They would probably throw an error except that your change returns from the sub before the arguments are used.
Please remove them to avoid someone looking at the code and then wondering the same thing that I just did.

rwstauner: Oh, after reading the pod for the sub I see that you were perhaps intending to emulate the `$perms` argument
in which case there are different problems:
`$file->module` is an array ref (see the previous test) not a module name, and the method accepts a hash **ref** so you'd need curly braces around the module/author pair, something like
`( { $file->module->[0]->{name} => $file->author } )`.
Problem there is that these "other" files are unlikely to ever have modules.

I think passing an empty hash ref would suffice: `( {} )`.
That way if the code goes past the "other files" check it won't error on not getting a hashref.
Files that are not "other files" should still get authorized then (which would be what you want in this case).

rwstauner:
I wasn't sure why we were using "authorized" for this until I read through #110 and found that exact suggestion.
I question whether it should be "indexed" or "authorized" that is used for this (see #323 for comparison).

It would be an easy swap.
Does any one have any thoughts on that?

andreeap:
I made the change in the test, but I am not sure what to do about the authorized/indexed issue. After I read #323 I understood better the difference between them and now  it makes more sense to set indexed => false.

oalders:
@rwstauner does this look good to you now?

rwstauner:
@oalders I was hoping you'd chime in on the "indexed vs authorized" question ;-)

oalders:
@rwstauner I just had another look #100.  You're correct. indexed => false is really what we want here.  We don't want these files to show up in searches, so that's clearly the right flag here.

andreeap:
I think this is ready if you can take a look at it.

oalders:
Looks good. @rwstauner ?

rwstauner:
The test passes because the file doesn't contain any content that would get it indexed.

The original issue (#110) mentioned a Makefile.PL with some (specific) pod, so that would be a good thing to test.

If you add something like this to your subtest:
```
    $file = new_file_doc( name => 'Makefile.PL', content => \<<EOF, );
=head1 NAME

Makefile.PL - blah

=head1 SYNOPSIS
EOF
    is( $file->indexed, 0, 'Makefile.PL with pod not indexed' );

```

it will expose a small (but familiar) bug in your code.

andreeap:
@rwstauner I did this mistake before, but again I didn't see it because of the test. I hope now it's better.

t/document/file.t
```diff
@@ -217,6 +217,26 @@ subtest 'Packages starting with underscore are not indexed' => sub {
...
+
+just a makefile description
+
+END
+    my $file = new_file_doc(
+        name       => 'Makefile.PL',
+        content_cb => sub { \$content }
+    );
+
+    $file->set_indexed( {} );
```
rwstauner: This lines isn't necessary, `new_file_doc` calls `set_indexed`.

rwstauner:
Glad to see you're learning ;-)

andreeap:
@rwstauner Thanks, I am glad too :)
I removed the line and please let me know if there is something else I could do.

oalders:
:+1:

PR #386, opened 2015-02-02
[WIP]skip broken files at indexing

andreeap:
Some tarballs contain broken symlinks or pipes that block the indexing process and by skipping those files, the indexer can run until completion.

### DIFF
--- a/cpanfile.snapshot
+++ b/cpanfile.snapshot
@@ -3367,6 +3367,21 @@ DISTRIBUTIONS
       ExtUtils::MakeMaker 6.59
       Test::More 0.96
       perl 5.008008
+  File-pushd-1.009
+    pathname: D/DA/DAGOLDEN/File-pushd-1.009.tar.gz
+    provides:
+      File::pushd 1.009
+    requirements:
+      Carp 0
+      Cwd 0
+      Exporter 0
+      ExtUtils::MakeMaker 6.17
+      File::Path 0
+      File::Spec 0
+      File::Temp 0
+      overload 0
+      strict 0
+      warnings 0
   Filesys-Notify-Simple-0.12
     pathname: M/MI/MIYAGAWA/Filesys-Notify-Simple-0.12.tar.gz
     provides:
@@ -7174,6 +7189,15 @@ DISTRIBUTIONS
       Test::Builder::Module 0
       Test::More 0.61
       perl 5.008_001
+  Test-RequiresInternet-0.04
+    pathname: M/MA/MALLEN/Test-RequiresInternet-0.04.tar.gz
+    provides:
+      Test::RequiresInternet 0.04
+    requirements:
+      ExtUtils::MakeMaker 0
+      Socket 0
+      strict 0
+      warnings 0
   Test-Routine-0.018
     pathname: R/RJ/RJBS/Test-Routine-0.018.tar.gz
     provides:
--- a/lib/MetaCPAN/Script/Release.pm
+++ b/lib/MetaCPAN/Script/Release.pm
@@ -200,6 +200,21 @@ sub run {
     $self->index->refresh;
 }
 
+sub _is_broken_file {
+    my $self = shift;
+    my $file = shift;
+
+    # the file is a pipe or the file doesn't exist
+    return 1 if ( -p $file || !-e $file );
+
+    # the file is a broken symbolic link
+    if ( -l $file ) {
+        my $syml = readlink $file;
+        return 1 if ( !-e $syml && !-l $syml );
+    }
+    return 0;
+}
+
 sub import_tarball {
     my ( $self, $tarball ) = @_;
     my $cpan = $self->index;
@@ -292,6 +307,7 @@ sub import_tarball {
                 = -d $File::Find::name
                 ? dir($File::Find::name)
                 : file($File::Find::name);
+            return if $self->_is_broken_file($File::Find::name);
             my $relative = $child->relative($tmpdir);
             my $stat     = do {
                 my $s = $child->stat;
--- a/t/script/release/skip_broken_files.t
+++ b/t/script/release/skip_broken_files.t
@@ -1,0 +1,57 @@
+use strict;
+use warnings;
+
+use Archive::Tar;
+use File::pushd;
+use FindBin;
+use HTTP::Request;
+use LWP::UserAgent;
+use MetaCPAN::Script::Release;
+use MetaCPAN::Script::Runner;
+use Test::More;
+use Test::RequiresInternet ( 'metacpan.org' => 443 );
+
+my $dir = tempd();
+
+my $ua = LWP::UserAgent->new(
+    agent      => 'metacpan',
+    env_proxy  => 1,
+    parse_head => 0,
+    timeout    => 30,
+);
+
+my $url
+    = 'https://cpan.metacpan.org/authors/id/S/SH/SHULL/karma-0.7.0.tar.gz';
+my $tarball = 'tarball.tar.gz';
+
+my $req = HTTP::Request->new( GET => $url );
+$req->header( Accept => '*/*' );
+my $res = $ua->request( $req, $tarball );
+
+my $arch = Archive::Tar->new;
+$arch->read($tarball);
+
+my $config = do {
+
+    # build_config expects test to be t/*.t
+    local $FindBin::RealBin = "$FindBin::RealBin/../..";
+    MetaCPAN::Script::Runner->build_config;
+};
+
+my $release = MetaCPAN::Script::Release->new($config);
+
+my $next = Archive::Tar->iter($tarball);
+while ( my $file = $next->() ) {
+    if ( $file->is_fifo ) {
+        ok(
+            $release->_is_broken_file($file),
+            $file->full_path . ' is a broken pipe'
+        );
+    }
+    elsif ( $file->is_symlink ) {
+        ok( $release->_is_broken_file($file),
+            $file->full_path . ' is a  broken symlink' );
+    }
+}
+
+done_testing();
--- a/cpanfile
+++ b/cpanfile
@@ -55,6 +55,7 @@ requires 'File::Basename';
 requires 'File::Find';
 requires 'File::Find::Rule';
 requires 'File::Find::Rule::Perl';
+requires 'File::pushd';
 requires 'File::Spec';
 requires 'File::Spec::Functions';
 requires 'File::Temp';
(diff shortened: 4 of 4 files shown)


oalders:
Are you able to add a test for this?

lib/MetaCPAN/Script/Release.pm
```diff
@@ -292,6 +306,7 @@ sub import_tarball {
                 = -d $File::Find::name
                 ? dir($File::Find::name)
                 : file($File::Find::name);
+            return if is_broken_file($File::Find::name);
```
oalders: `$self->_is_broken_file( $File::Find::name )`

lib/MetaCPAN/Script/Release.pm
```diff
@@ -200,6 +200,20 @@ sub run {
     $self->index->refresh;
 }
 
+sub is_broken_file {
```
oalders: `sub _is_broken_file`

lib/MetaCPAN/Script/Release.pm
```diff
@@ -200,6 +200,20 @@ sub run {
...
+    my $self = shift;
+
```
oalders: `my $file = shift;`

andreeap:
I don't have a test for this yet , but I want to add one.

t/script/release/skip_broken_files.t
```diff
@@ -0,0 +1,35 @@
+use strict;
+use warnings;
+
+use Archive::Tar;
+use HTTP::Request;
+use LWP::UserAgent;
+use MetaCPAN::Script::Release;
+use Test::More;
+
+#use Test::RequiresInternet ( 'https://metacpan.org/' => 443 );
```
oalders: `use Test::RequiresInternet ( 'metacpan.org' => 443 );`

See the SYNOPSIS https://metacpan.org/pod/Test::RequiresInternet#SYNOPSIS

cpanfile.snapshot
```diff
@@ -3839,6 +3839,19 @@ DISTRIBUTIONS
       IO::WrapTie::Slave 2.110
     requirements:
       ExtUtils::MakeMaker 0
+  IPC-Run-0.94
```
oalders: Do you know what pulled in IPC::Run?  I don't see it as a dep for Test::RequiresInternet and I don't see you using it anywhere else.

andreeap: I think this was from Archive::Tar::Wrapper. I use sublime as my editor and  when I switch branches cpanfile.snapshot keeps what I have installed before if I don't run carton install again.

andreeap:
This is ready for review.

t/script/release/skip_broken_files.t
```diff
@@ -0,0 +1,42 @@
+use strict;
+use warnings;
+
+use Archive::Tar;
+use HTTP::Request;
+use LWP::UserAgent;
+use MetaCPAN::Script::Release;
+use Test::More;
+
+use Test::RequiresInternet ( 'metacpan.org' => 443 );
```
oalders: This belongs with the other use statements above.

t/script/release/skip_broken_files.t
```diff
@@ -0,0 +1,42 @@
...
+
+my $ua = LWP::UserAgent->new(
+    parse_head => 0,
+    env_proxy  => 1,
+    agent      => 'metacpan',
+    timeout    => 30,
```
oalders: You can alpha sort these.

t/script/release/skip_broken_files.t
```diff
@@ -0,0 +1,42 @@
...
+);
+
+my $URL
+    = 'https://cpan.metacpan.org/authors/id/S/SH/SHULL/karma-0.7.0.tar.gz';
+my $tarball = 'tarball.tar.gz';
+
+my $req = HTTP::Request->new( GET => $URL );
+$req->header( Accept => '*/*' );
+my $res = $ua->request( $req, $tarball );
```
oalders: You can use a temporary directory for the download if this is going to end up in your checkout folder.  Look at https://metacpan.org/pod/File::Temp for that.

t/script/release/skip_broken_files.t
```diff
@@ -0,0 +1,42 @@
...
+
+my $arch = Archive::Tar->new;
+$arch->read($tarball);
+
+my $next = Archive::Tar->iter($tarball);
+while ( my $file = $next->() ) {
+    if ( $file->is_fifo ) {
+        my $skip = MetaCPAN::Script::Release->_is_broken_file($file);
+        ok( $skip eq 1, 'The file is a broken pipe' );
```
oalders: `ok( $skip, 'The file is a broken pipe' );`

t/script/release/skip_broken_files.t
```diff
@@ -0,0 +1,42 @@
...
+$req->header( Accept => '*/*' );
+my $res = $ua->request( $req, $tarball );
+
+my $arch = Archive::Tar->new;
+$arch->read($tarball);
+
+my $next = Archive::Tar->iter($tarball);
+while ( my $file = $next->() ) {
+    if ( $file->is_fifo ) {
+        my $skip = MetaCPAN::Script::Release->_is_broken_file($file);
```
oalders: You should be creating a Release object and then testing the method rather than calling this as a class method.
```
my $release = MetaCPAN::Script::Release->new( ... );
ok( $release->_is_broken_file( $file ), 'The file is a broken pipe');
```

t/script/release/skip_broken_files.t
```diff
@@ -0,0 +1,42 @@
...
+use Test::RequiresInternet ( 'metacpan.org' => 443 );
+
+my $ua = LWP::UserAgent->new(
+    parse_head => 0,
+    env_proxy  => 1,
+    agent      => 'metacpan',
+    timeout    => 30,
+);
+
+my $URL
```
oalders: `my $url`  Reserve upper case variable names for situations which require constants.

t/script/release/skip_broken_files.t
```diff
@@ -0,0 +1,63 @@
+use strict;
+use warnings;
+
+use Archive::Tar;
+use Cwd;
```
oalders: `use Cwd qw( cwd );`

Explicitly importing functions makes it easier to see why you imported this module without looking at the docs.

t/script/release/skip_broken_files.t
```diff
@@ -0,0 +1,63 @@
...
+    local $FindBin::RealBin = "$FindBin::RealBin/../..";
+    MetaCPAN::Script::Runner->build_config;
+};
+
+my $release = MetaCPAN::Script::Release->new($config);
+
+my $next = Archive::Tar->iter($tarball);
+while ( my $file = $next->() ) {
+    if ( $file->is_fifo ) {
+        ok( $release->_is_broken_file($file), 'The file is a broken pipe' );
```
oalders: The test output shows that there are 10 broken symlinks in this tarball.  Is this correct?  It might be helpful to include the file names in the test output so that we can eyeball this.

`ok( $release->_is_broken_file($file), $file->full_path . ' is a broken pipe' )`

You can do the same for symlinks.

t/script/release/skip_broken_files.t
```diff
@@ -0,0 +1,63 @@
...
+    }
+    elsif ( $file->is_symlink ) {
+        ok(
+            $release->_is_broken_file($file),
+            'The file is a  broken symlink'
+        );
+    }
+}
+
+chdir $original_dir;
```
oalders: I now realize I could have pushed you towards https://metacpan.org/pod/File::pushd which will create the temp dir and also take care of changing to the original directory, but this is fine.

andreeap:
I made the changes.

oalders:
I think this is ready to merge.  @rwstauner do you have any thoughts on this?  I asked @andreeap to use Test::RequiresInternet here as that's way easier than trying to make a fake dist containing the issues we're trying to deal with here.  I like this approach as it allows us to test more funky dists in future without having to a) move them to fakepan or 2) adding the tarball to the repository.

lib/MetaCPAN/Script/Release.pm
```diff
@@ -292,6 +307,7 @@ sub import_tarball {
                 = -d $File::Find::name
                 ? dir($File::Find::name)
                 : file($File::Find::name);
+            return if $self->_is_broken_file($File::Find::name);
```
rwstauner: Since you're not using the `$child` var that gets initialized above, you could move this return to the top
to avoid attempting to call `file()` on the bad value.

Because of the infrequency of these files it won't really be much of a performance gain, but it makes more sense than having it after.

rwstauner:
I find the File::pushd usage confusing (`$dir` is assigned, but not referenced again, it's not obvious that the rest of the file has been moved to a temp dir).

I'm also not a fan of redownloading the file every time the test suite runs.
Could we compromise and save the file to `t/var/tmp` so that it only downloads when it isn't already there?

rwstauner:
Does this need to be rebased now that the refactoring has been merged?

rwstauner:
While we do know ahead of time what the tarball should contain these tests are written in a way that there exists the potential for a bug.
If no matching files are found (for any reason), no tests will be run.
Currently this will cause the test suite to fail (because no tests were run)
but if somebody added an `ok()` anywhere else in the file, this test script could pass without having tested what you wanted to test.
Testing that the tests actually fired would be a good addition.
Using counter variables for each file type would be a simple way of double-checking.

oalders:
> Could we compromise and save the file to t/var/tmp so that it only downloads when it isn't already there?

We could use LWP's mirror() method to make it easy: https://metacpan.org/pod/LWP::UserAgent#ua-mirror-url-filename  Or we could use WWW::Mechanize::Cached as well.

andreeap:
The problem is that at least the method that skips the broken files was merged with the refactoring(the refactoring started as a way to improve the indexing process and I added there everything related to that) https://github.com/CPAN-API/cpan-api/blob/2c66076c5199de6deafb76615bc84293d876074a/lib/MetaCPAN/Model/Release.pm#L353, but there is not test for it. I can change this pull request to add just the test.

rwstauner:
Sounds good, thanks!

oalders:
@andreeap what's the status of this pull request?

ranguard:
Hi @andreeap I'm cleaning some stuff up, how's this going? Need any help?

oalders:
Looks like this branch has been abandoned?

oalders:
Closing for now.  We can re-open if there's future activity on this.

Issue #149, opened 2011-12-10
Attempt to determine "main module" for a release


rwstauner:
There are a lot of places where the "main module" for a dist would nice be to have.
The usual hack to get it is `$release =~ s/-/::/g`,
however we could try a little harder to guess it:
- Makefile.PL has ABSTRACT_FROM or VERSION_FROM
- dist.ini has main_module (or defaults to the `s///`)
- if the `s///` results in a module that doesn't exist, maybe just take the shortest/most common module namespace (for instance `LWP` or `Template`)

I don't know if there other/better alternatives, and many of these wouldn't always work,
but combining them (falling back to the `s///`) might work more often than any of them alone.

We could use the main_module for things like
- dist abstract #107
- more useful reverse dependencies link

oalders:
Fixed via #389

PR #389, opened 2015-02-10
add main_module attribute (adds new mappings)

andreeap:
This is a pull request for #149 , but it's still in progress. I also need to add a test.

*requires new mappings*

### DIFF
--- a/lib/MetaCPAN/Document/Release.pm
+++ b/lib/MetaCPAN/Document/Release.pm
@@ -217,6 +217,12 @@ has metadata => (
     source_only => 1,
 );
 
+has main_module => (
+    is       => 'rw',
+    isa      => 'Str',
+    required => 0,
+);
+
 sub _build_version_numified {
     return MetaCPAN::Util::numify_version( shift->version );
 }
--- a/lib/MetaCPAN/Model/Release.pm
+++ b/lib/MetaCPAN/Model/Release.pm
@@ -206,7 +206,45 @@ sub _build_document {
             ->put( { name => $self->distribution }, { create => 1 } );
     };
 
+    $self->_set_main_module( $self->modules, $document );
+
     return $document;
+}
+
+sub _set_main_module {
+    my $self = shift;
+    my ( $mod, $release ) = @_;
+
+    my @modules = @{$mod};
+
+    my $dist2module = $release->distribution;
+    $dist2module =~ s{-}{::}g;
+
+    if ( scalar @modules == 1 ) {
+
+        # there is only one module and it will become the main_module
+        $release->main_module( $modules[0]->module->[0]->name );
+        return;
+    }
+
+    foreach my $file (@modules) {
+
+        # the module has the exact name as the ditribution
+        if ( $file->module->[0]->name eq $dist2module ) {
+            $release->main_module( $file->module->[0]->name );
+            return;
+        }
+    }
+
+    # the distribution has modules on different levels
+    # the main_module is the first one with the minimum level
+    # or if they are on the same level, the one with the shortest name
+    my @sorted_modules = sort {
+        $a->level <=> $b->level
+            || length $a->module->[0]->name <=> length $b->module->[0]->name
+    } @modules;
+    $release->main_module( $sorted_modules[0]->module->[0]->name );
+
 }
 
 sub _build_files {
--- a/t/release/badpod.t
+++ b/t/release/badpod.t
@@ -7,12 +7,13 @@ use MetaCPAN::TestHelpers;
 
 test_release(
     {
-        name       => 'BadPod-0.01',
-        author     => 'MO',
-        authorized => \1,
-        first      => \1,
-        provides   => [ 'BadPod', ],
-        modules    => {
+        name        => 'BadPod-0.01',
+        author      => 'MO',
+        authorized  => \1,
+        first       => \1,
+        provides    => [ 'BadPod', ],
+        main_module => 'BadPod',
+        modules     => {
             'lib/BadPod.pm' => [
                 {
                     name             => 'BadPod',
--- a/t/release/binary-data.t
+++ b/t/release/binary-data.t
@@ -7,12 +7,13 @@ use MetaCPAN::TestHelpers;
 
 test_release(
     {
-        name       => 'Binary-Data-0.01',
-        author     => 'BORISNAT',
-        authorized => \1,
-        first      => \1,
-        provides   => [ 'Binary::Data', 'Binary::Data::WithPod', ],
-        modules    => {
+        name        => 'Binary-Data-0.01',
+        author      => 'BORISNAT',
+        authorized  => \1,
+        first       => \1,
+        provides    => [ 'Binary::Data', 'Binary::Data::WithPod', ],
+        main_module => 'Binary::Data',
+        modules     => {
             'lib/Binary/Data.pm' => [
                 {
                     name             => 'Binary::Data',
--- a/t/release/documentation-hide.t
+++ b/t/release/documentation-hide.t
@@ -17,6 +17,8 @@ is( $release->name, 'Documentation-Hide-0.01', 'name ok' );
 
 is( $release->author, 'MO', 'author ok' );
 
+is( $release->main_module, 'Documentation::Hide', 'main_module ok' );
+
 ok( $release->first, 'Release is first' );
 
 {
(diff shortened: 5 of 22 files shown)


lib/MetaCPAN/Script/Release.pm
```diff
@@ -490,6 +498,73 @@ sub import_tarball {
...
 
+sub find_main_module {
+    my $self = shift;
+    my ( $mod, $distribution ) = @_;
+
+    my @modules = @{$mod};
+
+    my $min_level          = 100;
+    my @distribution_parts = split /-/, $distribution;
+    my $distribution_name  = join '', @distribution_parts;
```
oalders: If you want to compare the distribution name to the module name, you could probably get a comparison variable without the split + join: 

```
my $dist2module = $distribution;
$dist2module =~ s{-}{::}g;
```

Now `HTML-Restrict` becomes `HTML::Restrict`.  If that matches a module, you've got your main_module.

lib/MetaCPAN/Script/Release.pm
```diff
@@ -490,6 +498,73 @@ sub import_tarball {
...
+
+    if ( scalar @modules eq 1 ) {
```
oalders: `if ( scalar @modules == 1 ) {`

You want a numeric comparison (not string).  Even if it actually works, using the `==` is less confusing.

lib/MetaCPAN/Script/Release.pm
```diff
@@ -490,6 +498,73 @@ sub import_tarball {
...
+
+        # there is only one module and it will become the main_module
+        my $main_module = shift @modules;
+        $self->main_module( $main_module->module );
+        return;
+    }
+    else {
```
oalders: Since you're returning in the `if` above, the `else` is not necessary.  If you get rid of it, you can save yourself one level of indentation.

lib/MetaCPAN/Script/Release.pm
```diff
@@ -490,6 +498,73 @@ sub import_tarball {
...
+        foreach my $file ( sort @modules ) {
+            my @module_parts = split /::/, $file->module->[0]->{name};
+            my $module_name = join '', @module_parts;
+
+            # the module has the exact name as the ditribution
+            if ( $module_name eq $distribution_name ) {
```
oalders: Using the code I suggested above: `if ( $file eq $dist2module ) {`

lib/MetaCPAN/Script/Release.pm
```diff
@@ -490,6 +498,73 @@ sub import_tarball {
...
+
+        # there is only one module and it will become the main_module
+        my $main_module = shift @modules;
+        $self->main_module( $main_module->module );
+        return;
+    }
+    else {
+        foreach my $file ( sort @modules ) {
+            my @module_parts = split /::/, $file->module->[0]->{name};
+            my $module_name = join '', @module_parts;
```
oalders: You can delete these two lines.

lib/MetaCPAN/Script/Release.pm
```diff
@@ -490,6 +498,73 @@ sub import_tarball {
...
+
+            # the module contains the ditribution name
+            if ( $module_name =~ /$distribution_name/ ) {
+                $self->main_module( $file->module );
+                return;
+            }
+
+            # the distribution has modules on different levels
+            # the main_module is the one with minimum level
+            if ( $file->level lt $min_level ) {
```
oalders: I'm not going to look at this logic too closely right now.  I think after you add some tests it will become clearer if this is the right approach.

lib/MetaCPAN/Script/Release.pm
```diff
@@ -490,6 +498,69 @@ sub import_tarball {
...
+
+    my $min_level   = 100;
+    my $dist2module = $distribution;
+    $dist2module =~ s{-}{::}g;
+
+    if ( scalar @modules == 1 ) {
+
+        # there is only one module and it will become the main_module
+        my $main_module = shift @modules;
+        $self->main_module( $main_module->module );
```
oalders: `$self->main_module( $modules[0] );`

andreeap:
Yes, I remember you said to write it like that, but I left it this way because otherwise it won't get into the is() statement and it will give me 'Can't call method "main_module" on unblessed reference at t/script/release/find_main_module.t line 27.' and I wanted to point out that I can't get the main_module value out of the release script.

rwstauner:
Why is `main_module` an attribute on the Script?  The script can process more than one release, and an attribute for the script is mainly intended to be passed in on the command line.

Can't we just assign the release's main_module attribute to the return value from a method call rather than caching it on the Script object first?
`find_main_module` could return the relevant value (instead of calling the attribute setter and then returning nothing)
and then:
`$release->main_module( $self->find_main_module( \@modules, $release->distribution ) );`

andreeap:
I modified the logic for choosing the main_module and also set it only for  $release without caching the value in the script. The main_module is set correctly for the example in the test, but I hardcoded the value because I am not sure how to access it there(I just take a tarball and pass it to the release script without doing something else). Should I rethink the test?

rwstauner:
Yes, let's rethink the test.
I see no reason that we need to explicitly call the Script... all we really want to test is that the release is saved with the expected main_module attribute.

We already have lots of tests in `t/release` that test our fake distributions.
Rather than writing up a custom test for this I'd suggest just adding "main_module" checks to our existing release tests.  Then we can add more fake releases if we need to test specific/edge cases rather than seeking out more "real" dists to have to download.
That will be more consistent with our existing test and much simpler (we won't need the `Test::RequiresInternet` and the boilerplate you had to figure out to run the Script).

We should be able to just remove the new test file you added (and the new dependencies in cpanfile\*)
and instead add a line like:
`main_module => 'BadPod',`
to files like `t/release/badpod.t`.

That being said, I tried that, and discovered some issues with the code that we should discuss.

My first question: Do we want "main_module" to be a module name like "Foo::Bar", or the "module" hashref like `{ name => "Foo::Bar", associated_pod => "" ... }`.
The latter is what it's being set to now (based on `MetaCPAN::Types::Internal::Module`), but the former is what it's testing for.
Originally I thought we would use just the name, but if anyone feels differently I'm open to discussion.

andreeap:
In order to test for the main_module, can I change this to only setting the name of the module or keep it the way it is now?

andreeap:
I set only the name of a module as the main_module, and when the releases from fakecpan are indexed, main_module is set, but when it gets to t/release/<release_name.t> this attribute is left undefined. What am I missing?

oalders:
> Originally I thought we would use just the name, but if anyone feels differently I'm open to discussion

I think we do just want to use the name rather than a HashRef here.  Good catch!

rwstauner:
@andreeap unfortunately that method (that whole file) is a bit confusing.  It could use some cleanup...
Primarily I see no obvious reason why there are multiple calls to `$release->put` (there are 3 within 20 lines of code), all inside if statements. That could surely be improved.
That's not something to fix in this pull request, however it's related to the issue.

You are calling `set_main_module` _after_ the last `put` so the document never makes it to the index.
In order to persist the change, I would suggest moving the call to `set_main_module` before the `$tmpdir->rmtree` line (before passing to the "latest" script), and then calling `$release->put` to make sure it is saved.

We can deal with cleaning up the confusion later (i think there are already outstanding branches about cleaning this up).

So, to clarify, I would suggest a change like this:

```diff
diff --git a/lib/MetaCPAN/Script/Release.pm b/lib/MetaCPAN/Script/Release.pm
index 02ef435..4f95d8c 100644
--- a/lib/MetaCPAN/Script/Release.pm
+++ b/lib/MetaCPAN/Script/Release.pm
@@ -481,16 +481,18 @@ sub import_tarball {
         $release->authorized(0);
         $release->put;
     }
 
+    $self->set_main_module( \@modules, $release );
+    $release->put;
+
     $tmpdir->rmtree;
 
     if ( $self->latest ) {
         local @ARGV = ( qw(latest --distribution), $release->distribution );
         MetaCPAN::Script::Runner->run;
     }
 
-    $self->set_main_module( \@modules, $release );
 }
 
 sub set_main_module {
     my $self = shift;

```

andreeap:
I rebased this pull request and added the field to some of the tests and it is ready to be reviewed.

oalders:
Looks good to me.  @rwstauner, any objections?  If not, I'll merge it tomorrow (Friday).

rwstauner:
@andreeap, thanks for the effort you put into updating all those tests.
This looks great!

Issue #299, opened 2014-02-26
Add 'changes_file' key to release data


rwstauner:
I made a comment to this effect when I added the changes controller:

https://github.com/CPAN-API/cpan-api/commit/e671469bf66b1a55d99bf69ecb2264adfeac0062#diff-4ed60210a2ea835de90db16374d32449R18

If we _did_ store the path to the change log on the release object
we could use that to fix CPAN-API/metacpan-web#1107.
We'd know ahead of time whether or not to make the link
and we wouldn't have to guess which file it was when trying to return it from the changes controller.

Note that this would require reindexing everything (#289) to be useful.

mickeyn:
most releases have this data, but it's not used in the changes controller - need to replace the hackish query to make use of it.

ranguard:
@mickeyn not sure what's the score with this ticket?

ranguard:
Just found it's needed for https://github.com/metacpan/metacpan-web/issues/1107

PR #392, opened 2015-02-18
add changes_file field (adds new mappings)

andreeap:
This is a fix for #299 and it is ready for review.

### DIFF
--- a/lib/MetaCPAN/Model/Release.pm
+++ b/lib/MetaCPAN/Model/Release.pm
@@ -208,6 +208,8 @@ sub _build_document {
 
     $self->_set_main_module( $self->modules, $document );
 
+    $document->changes_file( $self->get_changes_file( $self->files ) );
+
     return $document;
 }
 
@@ -248,6 +250,30 @@ sub _set_main_module {
     } @modules;
     $release->main_module( $sorted_modules[0]->module->[0]->name );
 
+}
+
+sub get_changes_file {
+    my $self          = shift;
+    my @files         = @{ $_[0] };
+    my @changes_files = qw(
+        Changelog
+        ChangeLog
+        CHANGELOG
+        Changes
+        CHANGES
+        NEWS
+    );
+
+    if ( $files[0]->distribution eq 'perl' ) {
+        foreach my $file (@files) {
+            if ( $file->name eq 'perldelta.pod' ) {
+                return $file->path;
+            }
+        }
+    }
+    foreach my $file (@files) {
+        return $file->path if grep { $_ eq $file->path } @changes_files;
+    }
 }
 
 sub _build_files {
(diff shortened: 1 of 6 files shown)


lib/MetaCPAN/Script/Release.pm
```diff
@@ -490,6 +492,25 @@ sub import_tarball {
     }
 }
 
+sub set_changes_file {
```
oalders: You'll just need to add some logic here to deal with `perl` distributions, since the Changes file there is `perldelta.pod`.  Have a look at the Changes controller.  Basically, once this work is complete and everything is reindexed, we'll want to use this new field rather than all of the logic in that controller.

So basically, you're taking the controller logic and applying it to the `changes_file` field so that we no longer have to discover where this file is on every request to the Changes controller.

andreeap:
And the field should have the name of the file or the path to the file?

oalders:
I think we want the path to the file.  In most cases it'll just be the file name *anyway*, but in Perl's case it could be something like `pod/perldelta.pd`

oalders:
@andreeap, there's a perl in our fakepan.  Could you test against that as well?

andreeap:
This is ready for review.

lib/MetaCPAN/Model/Release.pm
```diff
@@ -271,6 +273,34 @@ sub _build_files {
     return \@files;
 }
 
+sub set_changes_file {
```
oalders: Nitpick here.  This should be `get_changes_file`, since it's only returning the file name and not setting it on the document.  The setter is actually `$document->changes_file()`;

lib/MetaCPAN/Model/Release.pm
```diff
@@ -271,6 +273,34 @@ sub _build_files {
...
+        foreach my $file (@files) {
+            if ( $file->name eq 'perldelta.pod' ) {
+                return $file->path;
+            }
+
+        }
+    }
+
+    foreach my $file (@files) {
+        return $file->path if grep { $_ eq $file->name } @changes_files;
```
rwstauner: We're checking against the basename... do we want to allow these files to exist anywhere other than the root?
It's unlikely for these names to be anything else (other than the changes file)
but it just feels like a bug waiting to happen.

I think we want to test against `$file->path`.  Does anyone think we _should_ check just `$file->name`?
If so, we may want to consider something like collect all the possible candidates and then choose the one with the shortest path.  As one example, this could easily match the change log for any dist (or other project) embedded in the release.  Also something like `CPAN::Changes` probably has several files like this to use for test data.

Either way, it would be good to have one of our fakecpan dists include a file matching one of these names in a subdirectory so that we are intentional about how it works.

oalders: Good catch.  I think just in the top level, with the perl releases being the exception.  Adding a test case is a good idea.

andreeap:
I made the changes by adding a second file.

rwstauner:
Looks good, thanks!

oalders:
@andreeap could you rebase and fix the merge conflict?

PR #393, opened 2015-02-18
[WIP]Elasticsearch Upgrade

andreeap:
This is a pull request to test what happens when it's used the default version of elasticsearch that travis has.


.travis.yml
```diff
@@ -27,20 +27,6 @@ env:
     - DEPLOYMENT_PERL_VERSION=5.18
 
 
-before_install:
-  # We need to run a pre-1.0 instance of ES until we update everything.
-  - wget https://download.elasticsearch.org/elasticsearch/elasticsearch/elasticsearch-0.90.13.deb
-  - sudo dpkg -i --force-confdef elasticsearch-0.90.13.deb
-  - sudo service elasticsearch restart
-
```
oalders: I think we can keep everything below this line.  It's just the Elasticsearch install we don't need to downgrade.

oalders:
@andreeap could you add me as a contributor to this repository?

oalders:
Might be worth looking at https://metacpan.org/pod/Catalyst::Model::Search::ElasticSearch

oalders:
I've forked this and moved it to #398.  You should have write access to that branch.

PR #394, opened 2015-02-19
Turn off some PBP policies that aren't very helpful.

schwern:
I find the benefits of these policies to be arguable and negligible.
I spend more time making perlcritic happy than they provide for
readability or avoiding bugs.  If someone really finds `','` and `''`
hard to read we can turn them back on.

ProhibitInterpolationOfLiterals requires a special note because its
rationales are wrong.  Interpolated strings are compiled to literals
with concatenation and the compile time saved is wholly negligible
(especially for a project using Moose).

The policy cannot know if `'Foo $bar'` was intentional or not so the
rationale of communicating intent doesn't work.  On the other side,
strict will catch if `"Foo $bar"` is a mistake, but cannot catch if 
`'Foo $bar'` is a mistake and neither can be policy.

### DIFF
--- a/.perlcriticrc
+++ b/.perlcriticrc
@@ -13,18 +13,18 @@ verbose = 11
 [-Variables::ProhibitPunctuationVars]
 
 [CodeLayout::RequireTrailingCommas]
-severity = 5
+severity = 4
 
 [TestingAndDebugging::RequireUseStrict]
 equivalent_modules = Test::Routine
 
 [ValuesAndExpressions::ProhibitEmptyQuotes]
-severity = 5
+severity = 4
 
 [ValuesAndExpressions::ProhibitInterpolationOfLiterals]
 allow_if_string_contains_single_quote = 1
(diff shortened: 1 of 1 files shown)


oalders:
I actually do find these policies to be helpful (generally making intent clearer), but if you find they're getting in the way for you right now, I'd be fine with changing the severity rather than killing them entirely.  That way we could deal with them later on if we decide to be harsher.  You're pushing a lot of code right now, so I don't want this to get in the way of being productive.

schwern:
It didn't get in the way much so far, though I often just didn't run that test until I was ready to push.  It's going to be more of an annoyance when adapting Gitpan.

Lowering the severity required by the tests would be a good compromise.  Then the most nit-picky policies aren't required of every commit, but it's still policy and someone can do a severity 5 audit if they like.

oalders:
So, if you want to tweak the severity, we can merge that. :)

schwern:
Done.

PR #403, opened 2015-03-30
Actually use "indexed:false" logic for common files

rwstauner:
#385 set "indexed: false" for common release files (things that are not modules).
It had passing unit tests, however the code was actually not getting called during indexing:

* `set_indexed` is only called for module files
* the default was being set by `meta->should_index_file` before the object was created.

This adds an integration test and moves the "default" logic into a lazy builder where I think it makes the most sense.

Any thoughts?


oalders:
:+1:

PR #407, opened 2015-04-30
API-docs.md: fix all Explorer links to use https

dolmen:
The MetaCPAN Explorer currently doesn't work in Firefox with http://
Firefox blocks requests to http://api.metacpan.org as it is Cross-Origin. But they work in https://.
This is probably fixable server side (the Firefox console says to enable CORS), but in the mean time, just fix the links from the doc.


ranguard:
I've alto re-written http to https

PR #419, opened 2015-08-18
URI-escape rt bugs source url

rwstauner:
refs CPAN-API/metacpan-web#1577

### DIFF
--- a/lib/MetaCPAN/Script/Tickets.pm
+++ b/lib/MetaCPAN/Script/Tickets.pm
@@ -13,6 +13,7 @@ use Log::Contextual qw( :log :dlog );
 use Moose;
 use Parse::CSV;
 use Pithub;
+use URI::Escape qw(uri_escape);
 
 with 'MetaCPAN::Role::Script', 'MooseX::Getopt';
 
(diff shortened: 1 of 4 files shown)


lib/MetaCPAN/Script/Tickets.pm
```diff
@@ -203,6 +205,12 @@ sub parse_tsv {
     return \%summary;
 }
 
+sub rt_dist_url {
+    my ( $self, $dist ) = @_;
+    return 'https://rt.cpan.org/Public/Dist/Display.html?Name='
+        . uri_escape($dist);
+}
+
```
oalders: We could also have dome something like:

```
    use URI;
    use URI::QueryParam;

    my $uri = URI->new( 'https://rt.cpan.org/Public/Dist/Display.html' );
    $uri->query_param( Name => $dist );
    return $uri;
```

That would also handle the escaping, but TIMTOWDI and it's only one URL param.  Thought I'd point it out, though, since I've found `URI` to be quite handy.

rwstauner: Good point.  That does seem like a better way to do it.

Issue #422, opened 2015-09-11
Pod renderer needs to be smarter about non-latest pod links


rwstauner:
Examples:
- https://metacpan.org/pod/release/JROBINSON/DBIx-Class-Manual-SQLHackers-1.2/lib/DBIx/Class/Manual/SQLHackers.pod
  a pod file (no package (therefore no 02packages)) links to other pod files (ditto)
- https://github.com/CPAN-API/metacpan-web/issues/1567
  a release should probably link to other enclosed pages in the same version

For these kinds of things we either need to make the `/pod` endpoint smart about where someone is linking from (which seems like a terrible idea) or we need to hook into the renderer and determine what the base for a link should be:
- if destination is not indexed: _full link_
- if link is to separate dist: _short link_ (not indexed would be handled by first item)
- if release is not latest (link is internal): _full link_ (keep to specific version)
- if release is latest: _short link_

Could we simplify the logic by saying "anything in this dist gets a full link, anything outside this dist gets the short link"?  The change there would be in not linking people to the canonical (latest) url when maybe we could, so the question becomes: _is that better than linking to the latest url when we shouldn't_? I think this is essentially what sco is doing.

Thoughts?

ribasushi:
> anything in this dist gets a full link, anything outside this dist gets the short link

:+1:

If this gets implemented things become much much better for a production end-user trying to read the docs for their present version (and lets face it - production running on the latest version of something CPAN is very very very rare)

haarg:
My preference would be to keep short links for in-dist modules when viewing a page using the short link.  I also don't think we should link to un-indexed files that aren't in the dist.

I've written up a preliminary idea for how to handle this: https://github.com/haarg/cpan-api/commit/ba0e35f80f3cb6517016a69d130c73cd0bbec4b4  It is completely untested at this point.  It adds a query parameter to the pod end point, which if present will use full links for anything in the same dist.  The front end can then use that parameter when viewing a page using a short link.

rwstauner:
I suspect that someone (like ranguard) would vote for short (canonical) urls as they are good for caching and search indexing.

Using the short links when viewing a short link would perhaps accomplish both objectives.  It's obviously more complicated, but should be doable.  I haven't had a chance to look over that commit yet.

ranguard:
As long as we have short links (linking to latest version), and this is what other dists link to, and is also what is set as canonical in the HTML it shouldn't affect SEO/caching having 'in dist' links localised to a version.

[...]

PR #423, opened 2015-09-14
Improve pod links to files in the same dist

haarg:
For files inside the same dist, improve the links generated in the pod.  For files that aren't modules, use /pod/distribution/$dist/$file.  For unauthorized modules or when the permalinks request parameter is used, use /pod/release/$author/$release/$file.

This does not yet address when multiple files have the same pod name.


haarg:
This is the API side of my idea to address #422.

oalders:
Do we need to add any tests here?  Do you have any URLs you can suggest we try looking at with this branch?

oalders:
Ah, I see we can check URLs in the referenced issue.

PR #425, opened 2015-10-19
match 'other files' by path, not filename

haarg:
The list of files given as 'other' are meant to be excluded from
indexing in the root of a dist.  Some may intentionally exist in
subdirectories (particularly in lib/) and meant to be indexed.  Instead
of matching based on the filename, match based on the full path so it
only applies to files in the root directory (so far).

Fixes CPAN-API/metacpan-web#1576


oalders:
:+1:

PR #430, opened 2016-01-31
Purge fastly by author and module name

ranguard:


lib/MetaCPAN/Role/Fastly.pm
```diff
@@ -0,0 +1,249 @@
...
+# Make sure the CDN NEVER caches, ignore any other cdn_cache_ttl settings
+has 'cdn_never_cache' => (
+    is      => 'rw',
+    isa     => Bool,
+    default => 0,
+);
+
+has 'browser_max_age' => (
+    is      => 'rw',
+    isa     => 'Maybe[Int]',
```
oalders: Should use imported types.

lib/MetaCPAN/Role/Fastly.pm
```diff
@@ -0,0 +1,249 @@
...
+=head2 $c->cdn_times;
+
+Returns a hashref of 'one_hour', 'one_day', 'one_week'
+and 'one_year' so we don't have numbers all over the place
+
+=cut
+
+## Stuff for working with Fastly CDN
+
+has '_surrogate_keys' => (
```
oalders: Don't need to quote attributes, but don't need to change this.

oalders:
Looks good to me.  :+1:

PR #433, opened 2016-03-07
fix 'my...if...' trap

mickeyn:
`my ... if ...` can only lead to trouble :)


oalders:
Thanks!  Looks good.

The test failures are unrelated to this branch, but I'll clean them up before we merge this.

oalders:
Can't automatically merge because of failing checks, so I cherry-picked the commit into https://github.com/CPAN-API/cpan-api/pull/434

PR #445, opened 2016-04-06
[WIP] some hacks to allow indexing on ES 2.3

mickeyn:
Some of the hacks used (except for local configuration files) to allow indexing on ES 2.3
I assume the indexing is not complete and this PR is only meant for the discussion and change reference for actual future changes.


lib/MetaCPAN/Script/Author.pm
```diff
@@ -112,21 +112,21 @@ sub author_config {
         log_debug {"Skipping $pauseid (newer version in index)"};
         return undef;
     }
-    my $json = $file->slurp;
-    my $author = eval { JSON::XS->new->utf8->relaxed->decode($json) };
 
-    if ($@) {
+    my $author;
```
oalders: Would be nice to start using `Try::Tiny` here and `Cpanel::JSON::XS`

lib/MetaCPAN/Server/QuerySanitizer.pm
```diff
@@ -50,10 +50,13 @@ sub _scan_hash_tree {
     my $ref = ref($struct);
     if ( $ref eq 'HASH' ) {
         while ( my ( $k, $v ) = each %$struct ) {
-            if ( $k eq $key ) {
-                MetaCPAN::Server::QuerySanitizer::Error->throw(
-                    message => qq[Parameter "$key" not allowed], );
-            }
+            # Mickey: disabling this check for 'script' key since
+            #         for ES 1.7 I need to use it in the
+            #         function_score syntax
```
oalders: Could you elaborate on why this needs to be done?  Internally we can certainly use `script` but allowing it via the API would expose us to exploits.

lib/MetaCPAN/Document/Favorite.pm
```diff
@@ -41,7 +41,7 @@ Sets the C<_timestamp> field to the value of L</date>.
 
 has timestamp => (
     is        => 'ro',
-    timestamp => { path => 'date', store => 1 },
+    timestamp => {}, # { path => 'date', store => 1 },
```
oalders: This is good.  We'd just need to add the timestamp manually.  If we don't do that in this pull request, we should open an issue for it so that we don't forget.

lib/MetaCPAN/Server/QuerySanitizer.pm
mickeyn: one of the early hacks. couldn't make the `prefer_shorter_module_names_400` to work as given script name... I translated it to be used through the new `function_score` syntax, but that required a `script` key.
I'll try again to make the original stored script to work and restore the restriction.

lib/MetaCPAN/Document/Favorite.pm
mickeyn: Also to consider: `_timestamp` seems to be depracated in ES2+
See https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-timestamp-field.html

lib/MetaCPAN/Server/QuerySanitizer.pm
mickeyn: restored in 97c61e0

note that in ES2 we need to add the following configuration to `elasticsearch.ym`:
```
script.inline: on
```

mickeyn:
992cb30 needs https://github.com/CPAN-API/metacpan-puppet/pull/98 and https://github.com/CPAN-API/metacpan-puppet/issues/96 to be resolved.

oalders:
Let's merge this in!

PR #449, opened 2016-04-15
updated snapshot

ranguard:


ranguard:
This was done in mi/es2 - so closing this

PR #451, opened 2016-04-17
mickey/es2 no lazy build

mickeyn:


ranguard:
@oalders this looks good to me, but as travis is failing from mi/es2 I'm nervous in merging code I don't really know so will leave with you

lib/MetaCPAN/Server/Model/CPAN.pm
```diff
@@ -11,9 +11,9 @@ extends 'Catalyst::Model';
 with 'CatalystX::Component::Traits';
 
 has esx_model => (
-    is         => 'ro',
-    lazy_build => 1,
-    handles    => ['es'],
+    is      => 'ro',
+    handles => ['es'],
+    default => sub { MetaCPAN::Model->new( es => $_[0]->servers ) },
```
oalders: Any reason not to make this `lazy` as well?

lib/MetaCPAN/Document/File.pm
```diff
@@ -427,12 +427,12 @@ ArrayRef of ArrayRefs of offset and length of pod blocks. Example:
...
 
 has pod_lines => (
-    is         => 'ro',
-    required   => 1,
-    isa        => 'ArrayRef',
-    type       => 'integer',
-    lazy_build => 1,
-    index      => 'no',
+    is      => 'ro',
+    isa     => 'ArrayRef',
```
oalders: I think we should make sure to use our own type library wherever possible and just import the types we're using.

```
use MetaCPAN::Types qw( ArrayRef );

...
isa => ArrayRef,
...
```

lib/MetaCPAN/Document/File.pm
```diff
@@ -527,24 +527,24 @@ has version => (
...
 
 =cut
 
 has version_numified => (
-    is         => 'ro',
-    isa        => 'Str',
-    lazy_build => 1,
-    required   => 1,
+    is      => 'ro',
+    isa     => 'Num',
```
oalders: Use imported type here.

lib/MetaCPAN/Document/File.pm
```diff
@@ -581,11 +581,11 @@ sub _build_path {
 }
 
 has dir => (
-    is         => 'ro',
-    lazy_build => 1,
-    isa        => 'Str',
-    required   => 1,
-    index      => 'not_analyzed'
+    is      => 'ro',
+    isa     => 'Str',
```
oalders: Imported type.

lib/MetaCPAN/Document/Release.pm
```diff
@@ -142,10 +142,10 @@ has [qw(distribution name)] => (
 );
 
 has version_numified => (
-    is         => 'ro',
-    required   => 1,
-    isa        => 'Str',
-    lazy_build => 1,
+    is      => 'ro',
+    isa     => 'Num',
```
oalders: Imported type

lib/MetaCPAN/Model/User/Account.pm
```diff
@@ -90,11 +90,11 @@ is true if the user is connected to a PAUSE account or he L</passed_captcha>.
 =cut
 
 has looks_human => (
-    is         => 'ro',
-    isa        => 'Bool',
-    required   => 1,
-    lazy_build => 1,
-    clearer    => 'clear_looks_human',
+    is      => 'ro',
+    isa     => 'Bool',
```
oalders: Imported type

lib/MetaCPAN/Script/ReindexDist.pm
```diff
@@ -10,9 +10,10 @@ use Moose;
 with 'MetaCPAN::Role::Script', 'MooseX::Getopt';
 
 has distribution => (
-    is         => 'ro',
-    isa        => 'Str',
-    lazy_build => 1,
+    is      => 'ro',
+    isa     => 'Str',
```
oalders: Imported type

lib/MetaCPAN/Script/ReindexDist.pm
```diff
@@ -24,9 +25,10 @@ sub _build_distribution {
 }
 
 has releases => (
-    is         => 'ro',
-    isa        => 'ArrayRef',
-    lazy_build => 1,
+    is      => 'ro',
+    isa     => 'ArrayRef',
```
oalders: Imported type

lib/MetaCPAN/Script/ReindexDist.pm
```diff
@@ -40,9 +42,10 @@ sub _build_releases {
 }
 
 has sources => (
-    is         => 'ro',
-    isa        => 'ArrayRef',
-    lazy_build => 1,
+    is      => 'ro',
+    isa     => 'ArrayRef',
```
oalders: Imported type

lib/MetaCPAN/Server/Diff.pm
```diff
@@ -18,14 +18,16 @@ has [qw(source target)] => (
...
+    lazy    => 1,
+    builder => '_build_raw',
 );
 
 has structured => (
-    is         => 'ro',
-    isa        => 'ArrayRef',
-    lazy_build => 1,
+    is      => 'ro',
+    isa     => 'ArrayRef',
```
oalders: Imported type

lib/MetaCPAN/Server/Model/CPAN.pm
mickeyn: It's accessed in BUILD... didn't seem like something that really needs to be lazy.
What's the benefit of it being lazy?

lib/MetaCPAN/Document/File.pm
```diff
@@ -486,17 +486,20 @@ Source Lines of Pod. Returns the number of pod lines using L</pod_lines>.
 =cut
 
 has slop => (
-    is         => 'ro',
-    required   => 1,
-    isa        => 'Int',
-    is         => 'rw',
-    lazy_build => 1,
+    is      => 'ro',
+    isa     => 'Int',
```
oalders: Imported type

lib/MetaCPAN/Document/File.pm
```diff
@@ -486,17 +486,20 @@ Source Lines of Pod. Returns the number of pod lines using L</pod_lines>.
...
+    lazy    => 1,
+    default => '_build_slop',
+    writer  => '_set_slop',
```
oalders: 👍

lib/MetaCPAN/Document/File.pm
```diff
@@ -486,17 +486,20 @@ Source Lines of Pod. Returns the number of pod lines using L</pod_lines>.
...
 );
 
 sub _build_slop {
     my $self = shift;
     return 0 unless ( $self->is_perl_file );
     $self->_build_pod_lines;
+
+    # danger! infinite recursion if not set by `_build_pod_lines`
+    # we should probably find a better solution -- Mickey
     return $self->slop;
```
oalders: Good to note this.  :)

lib/MetaCPAN/Document/Module.pm
```diff
@@ -104,11 +97,11 @@ has associated_pod => (
...
-sub _build_version_numified {
-    my $self = shift;
-    return 0 unless ( $self->version );
-    return MetaCPAN::Util::numify_version( $self->version ) . q{};
-}
+# sub _build_version_numified {
+#     my $self = shift;
+#     return 0 unless ( $self->version );
+#     return MetaCPAN::Util::numify_version( $self->version ) . q{};
+# }
```
oalders: Let's just delete this.  If we need to find it, we can look for this commit.

oalders:
Just had a few comments that would  be nice to address while we're already looking at this.  Looks good!  It really bugs me that the Travis build is failing.  Will try to sort that next.

mickeyn:
OK, doing a `MetaCPAN::Types` sweep, just checking things are still working and will push shortly.

lib/MetaCPAN/Server/Model/CPAN.pm
oalders: In general it would be my preference if everything that required a builder or a default with an anonymous sub were lazy, unless there's a compelling reason for that not to be the case.  Mostly because that gives us more predictability around exactly when things happen.  We could well move steps out of `BUILD` in future and then weird things might happen here.  Or, if something gets added before this in the BUILD that messes with the order of operations we have something else to debug.

So, in general, if the object does the least possible amount of work before anything really needs to happen, that seems like the easier situation to debug.  At the very least, it won't make the logic any _worse_.  :)

mickeyn: I already reverted (1a7a689) and fixed this one (259f6ef), as it broke the web server startup

mickeyn:
@oalders added 9e491b8 ... doesn't seem to add new breakage :)

lib/MetaCPAN/Server/Model/CPAN.pm
```diff
@@ -11,9 +11,9 @@ extends 'Catalyst::Model';
 with 'CatalystX::Component::Traits';
 
 has esx_model => (
-    is      => 'ro',
-    handles => ['es'],
-    default => sub { MetaCPAN::Model->new( es => $_[0]->servers ) },
+    is         => 'ro',
+    lazy_build => 1,
```
oalders: This can still be `lazy => 1` with a default, but let's merge this now and we can look at it later.

oalders:
Thanks very much!

lib/MetaCPAN/Server/Model/CPAN.pm
mickeyn: this one is probably not relevant anymore. there are no mentions of
`lazy_build` in the code now.

PR #462, opened 2016-04-23
Added script which fetches and indexes river data

### LINKED ISSUE #460: Import river position data and make available via the MetaCPAN API
In the River discussion at the QAH 2016 we agreed one of the things that would help most is if a dist's river position were available via the MetaCPAN API, and also displayed on a dist's home page on MetaCPAN (I'll raise a separate ticket for that).

I'm calculating river position for all dists every week, and have agreed to put this data somewhere. After a chat with @oalders we agreed with that initially this will be simple JSON data like the following:

    [
        {
            "dist":      "System-Command",
            "total":     92,
            "immediate": 4,
            "bucket",    2
        },
        {
            "d": "Text-Markdown",
            "t": 92,
            "i": 56,
            "b", 2
        }
    ]

Here's what those fields are:

 * **immediate** is the number of immediate downstream dependents (considering *required*, *non-developer* prereqs) that the dist has
 * **total** is the total number of downstream dependents
 * **bucket** is a number between 0 and 5, the logarithmic binning of *total*. 0 means no downstream deps, and 5 is the head of the river.

JSON always seems stupidly verbose, so for internal things I tend to use shorter names, like the second example above. Or would you rather go with verbosity?

Once we've agreed on the above format, I'll decide where to publish it so you can grab a first version, then I'll set up something to ensure it's getting regularly updated.

jberger:
closes #460

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@
 /var
 /t/var/tmp/
 /t/var/darkpan/
+/t/var/log/
 /etc/metacpan_local.pl
 metacpan_server_local.conf
 
--- a/lib/MetaCPAN/Document/Distribution.pm
+++ b/lib/MetaCPAN/Document/Distribution.pm
@@ -7,7 +7,7 @@ use namespace::autoclean;
 use Moose;
 use ElasticSearchX::Model::Document;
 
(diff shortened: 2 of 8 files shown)


oalders:
Thanks very much!

PR #465, opened 2016-04-27
use scratch disk if it exists

ranguard:


lib/MetaCPAN/Model/Archive.pm
```diff
@@ -77,7 +77,11 @@ has _tempdir => (
     init_arg => undef,
     lazy     => 1,
     default  => sub {
-        return File::Temp->newdir;
+
+        my $scratch_disk = '/mnt/scratch_disk';
```
jberger: I always wonder about hard-coded paths in lexicals like this. Could/should this be in configuration somewhere or else at least a package global?

ranguard: Yes, agree 100%, but my role is JFDI so we get migrated to ES 2...  and indexing is taking too long.

Please do send a better PR with the data configured from metacpan_server.conf (I'll handle the metacpan_server_local.conf to match via puppet so it's specific to the server) :)

oalders: We are starting to be better about hard-coded paths.  In this case, I think the path will be the same on the Vagrant box, but this will probably explode under Travis, so we at least need to special case that.

oalders: I misread this.  Should be fine on Travis.  We can fix the paths as part of a cleanup later.

shadowcat-mst:
I said to use an env var for this, and then set it in the minion script.

Why did we change it to be hard coded instead?

-- 
Matt S Trout - Shadowcat Systems - Perl consulting with a commit bit and a clue

http://shadowcat.co.uk/blog/matt-s-trout/   http://twitter.com/shadowcat_mst/

Email me now on mst (at) shadowcat.co.uk and let's chat about how our CPAN
commercial support, training and consultancy packages could help your team.

PR #487, opened 2016-05-27
t/model/archive: replace size check with content check

mickeyn:
hopefully make the test pass on 5.22 on travis :)


mickeyn:
all passes on 5.22

oalders:
👍

PR #489, opened 2016-06-02
Support queueing from Script::Release

mickeyn:
support `--queue` flag in `release` script, to allow it to queue its jobs.

it will allow speeding up crons like:
`... bin/metacpan release --skip --age 25 --detect_backpan --latest `
by sending all their jobs to the minion queue using:
`... bin/metacpan release --queue --skip --age 25 --detect_backpan --latest `


oalders:
Interesting approach.  :)

Issue #466, opened 2016-04-29
V1 not machine v0


ranguard:
https://api-v1.metacpan.org/module/OALDERS/ElasticSearchX-Model-1.0.0/lib/ElasticSearchX/Model/Bulk.pm

vs

https://api.metacpan.org/module/OALDERS/ElasticSearchX-Model-1.0.0/lib/ElasticSearchX/Model/Bulk.pm

ranguard:
```perl
use strict;
use warnings;

use Test::More;
use JSON;
use Test::Deep;
use HTTP::Tiny;

my @paths = qw(
	/module/ISHIGAKI/Archive-Any-Lite-0.11/lib/Archive/Any/Lite.pm
);

my $old_stub = 'https://api.metacpan.org/';
my $new_stub = 'https://api-v1.metacpan.org/';

for my $path (@paths) {
	my $old_res = HTTP::Tiny->new->get($old_stub . $path);
	my $new_res = HTTP::Tiny->new->get($new_stub . $path);

	my $old = decode_json($old_res->{content});
	my $new = decode_json($new_res->{content});

	cmp_deeply(
	  $new,
	  $old,
	  "$path matches"
	);

}

done_testing();
```

ranguard:
FYI for the above:
```
not ok 1 - /module/ISHIGAKI/Archive-Any-Lite-0.11/lib/Archive/Any/Lite.pm matches
#   Failed test '/module/ISHIGAKI/Archive-Any-Lite-0.11/lib/Archive/Any/Lite.pm matches'
#   at cmp.pl line 23.
# Comparing hash keys of $data
# Missing: 'abstract', 'description', 'level', 'version_numified'
# Extra: 'download_url', 'section'
1..1
```

mickeyn:
#490 will leave us with 1 extra field in V1 - `download_url`, qn: do we care enough to remove it from the mapping or do we need it?

oalders:
I think we need `download_url`.  It's quite handy.

mickeyn:
great so with #490 merged that's the only diff we're gonna have after full-reindex (need to change mapping again).

mickeyn:
reindex is done. @ranguard: is this still an issue?

ranguard:
@mickeyn I've updated the code example above - all looking good - I'll close this ticket but maybe you could put this code somewhere for general testing?

PR #490, opened 2016-06-02
update mapping

mickeyn:
added missing `version_numified` and removed internal `section`

this change is to fix most of mismatches reported by the script in #466


oalders:
👍

PR #492, opened 2016-06-06
remove gid & uid from file stat

mickeyn:
from IRC:
```
20:39:06 � ranguard� any one know why gid and uid is in {stat}, other than someone not cleaning it out - e.g. 
                     https://api.metacpan.org/module/OALDERS/ElasticSearchX-Model-1.0.0/lib/ElasticSearchX/Model/Bulk.pm 

01:47:22 � oalders� ranguard: good question. no reason to keep that info
```


oalders:
👍

PR #497, opened 2016-06-19
improve link choices for in-dist modules/pod

haarg:
This is an improvement over my last attempt of providing in-dist links for non-indexed or non-latest pod.


oalders:
@mickeyn would you like to review/merge this?

mickeyn:
sure. we should probably want to start thinking of moving those queries to the API side, but other than that it seems fine.

PR #501, opened 2016-07-04
overhaul Script/Backpan

mickeyn:
many changes to make it useful again (but hopefully we won't need to use it, at least not often)

supports `undo` to move `backpan` status back to `cpan` and `files-only` to update `file` index only.

another change is in Script/Release - rename the cpan files list attribute - it is the opposite of backpan list (as the name suggested)

### DIFF
--- a/lib/MetaCPAN/Script/Backpan.pm
+++ b/lib/MetaCPAN/Script/Backpan.pm
@@ -3,66 +3,228 @@ package MetaCPAN::Script::Backpan;
 use strict;
 use warnings;
 
-use BackPAN::Index;
 use Moose;
 
+use Log::Contextual qw( :log :dlog );
+use BackPAN::Index;
+use MetaCPAN::Types qw( Bool HashRef Str );
+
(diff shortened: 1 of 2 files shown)


lib/MetaCPAN/Script/Backpan.pm
```diff
@@ -3,66 +3,228 @@ package MetaCPAN::Script::Backpan;
...
+    unless ( -e $ls ) {
+        log_error {"File $ls does not exist"};
+        exit;
+    }
+    log_info {"Reading $ls"};
+    my $cpan = {};
+    open my $fh, "<:gzip", $ls;
+    while (<$fh>) {
+        my $path = ( split(/\s+/) )[-1];
+        next unless ( $path =~ /^authors\/id\/\w+\/\w+\/(\w+)\/(.*)$/ );
```
oalders: Would be nice to use a different regex delimiter here so we don't need to escape the slashes, eg `m{...}`

mickeyn: yeah, that was copy-pasted from Script/Release... I'll update both files.

Issue #502, opened 2016-07-06
Include number of {BackPAN,CPAN} {releases,distributions} on authors


neilb:
It would be handy if we could add the following information to authors:

 * number of distributions the author has on CPAN
 * number of distributions the author has on BackPAN

Then on the web interface, if an author has no distributions on CPAN, but has dists on BackPAN, then the author page could reflect that, instead of just saying "no releases" (that change to the web interface was suggested by @book in [metacpan-web #1692](https://github.com/metacpan/metacpan-web/issues/1692))

mickeyn:
1. a distribution is not identified by an author, so let's stick to 'releases'.
2. do you want to aggregate the status 'latest' into 'cpan' or keep them separated?

my idea is to add to the author end-point (the API side) another key (we can choose another name), and the endpoint will return something like:
```
{
   "pauseid" : "NEILB",
   "name" : "Neil Bowers",
   ...
   "release_count": {
      "cpan": 87,
      "latest": 117,
      "backpan": 484
   }
}
```

WEB side changes can follow once this data is in the index & served through the author endpoint

neilb:
@mickeyn with the following interpretations?

 * **latest** means they have that many releases that are the latest of the relevant distribution in their CPAN author directory
 * **cpan** means they have that many additional releases sitting in their author directory
 * **backpan** is the total number of releases in their BackPAN directory. So **backpan** includes **latest** and **cpan**.

mickeyn:
A small correction: **backpan** (MetaCPAN's status=**backpan**) cannot include **latest** as it is built of the list of files not found in `CPAN/indices/find-ls.gz` - so for us that's a count of **`in BackPAN but not in CPAN`**.

neilb:
Ah, right. So could we have **backpan-only** and **backpan-total** as two separate keys, which would make it clear? But if that would be a pain, then the three keys you listed would be handy anyway.

mickeyn:
sure, but that means I aggregate the numbers of the 3 statuses we have to get **backpan-total** - it's not the real thing

neilb:
yeah, could drop the total, but maybe change the **backpan** key to **backpan-only**, to make it clearer what it is?

mickeyn:
sure

neilb:
Cool. I like your proposal.

mickeyn:
now that we have the numbers, do we want to change the wording of the
"No... were found" line (what would you suggest to write?) and/or show count match near the links?
something like -
[...]

PR #504, opened 2016-07-08
GET controller for Author, add release count info

mickeyn:
Per @neilb's request #502 

This is done in the controller.

After playing with this solution *and* with adding the data to the index, I think this is the better one as it will show data in sync with the release index (no need for the Author script to update the numbers) and won't require remapping of the index.

Still, open for comments / suggestions.


oalders:
Could we add a test that checks to see if this new data structure exists in the author data?

mickeyn:
sure

mickeyn:
done :)

oalders:
Thanks!

neilb:
Thanks @mickeyn!

PR #505, opened 2016-07-11
Add author links to author handler response

mickeyn:
This will allow us:
1. Serve the links as extra info from the API as well as the WEB.
2. Not concatenate those links using code within the WEB templates.
3. Reuse the links in templates.

Moving functionality from WEB to API.
This was done to allow cleaning up some template code and reuse links.


oalders:
Nice!

PR #506, opened 2016-07-11
Add url_prefix option to pod endpoint

haarg:
This allows an alternate URL prefix to be used when rendering Pod.


oalders:
Nice!

PR #507, opened 2016-07-27
mickey/2 birds

mickeyn:
this PR fixes 2 issues (that are very much related):

1. corrects the way --queue is handled by `Script::Release` so `--latest` is not ignored
2. fixes the `latest` bit overrun by the last `document->put` in Script::Release which screws up so much in the data - we got `latest` to be applied to the `file` index only without `release` being updated (the reason for many Web issues)


oalders:
Thanks!

PR #508, opened 2016-08-01
script/mapping: use the provided --delete flag

mickeyn:
we currently ignore this flag which is bad if we want to add new functionality to the script without destroying the mapping.


oalders:
Thanks!

PR #510, opened 2016-08-10
script/latest: support queueing + code reuse

mickeyn:
1. Moved queueing infra. to Role::Script for reuse.
2. Added support for queueing in Script::Latest
   (will be useful for later features)

This is part of creating a 'missing latest release' checker, but will also allow extending the use of the queue and make it easily available from all scripts.


mickeyn:
the failing test is login/openid... nothing to do with this script change.

oalders:
Right.  Not sure why the URL has changed.  This should be unrelated to the warnings from that module.  https://travis-ci.org/metacpan/metacpan-api/builds/151138603#L2746

PR #512, opened 2016-08-14
server: use to_app

mickeyn:
any reason this added function wasn't used?
the current code relying on the returned value of `require` is not explicit and not as clear.


mickeyn:
d'oh. forgot to run the tests.

oalders:
I think I added this function at QAH for use in tests.  Good idea to use it outside of tests as well.

PR #516, opened 2016-08-30
[WIP] distribution: bugs -> rt + github

mickeyn:
example entry:

``` json
      {
        "_type": "distribution",
        "_source": {
          "bugs": {
            "github": {
              "active": 10,
              "source": "https://github.com/ranguard/text-vcard",
              "open": 10,
              "closed": 30
            },
            "rt": {
              "closed": "6",
              "patched": 0,
              "active": "2",
              "resolved": 6,
              "rejected": 0,
              "open": 0,
              "new": 2,
              "source": "https://rt.cpan.org/Public/Dist/Display.html?Name=Text-vCard",
              "stalled": 0
            }
          },
          "name": "Text-vCard"
        },
        "_score": 1,
        "_index": "cpan_v1",
        "_id": "Text-vCard"
      }
```

WIP since after agreeing on this we will need to reindex the `distribution` type - which means deleting it, remapping and rebuilding using the tickets and river scripts.

### DIFF
--- a/lib/MetaCPAN/Types/Internal.pm
+++ b/lib/MetaCPAN/Types/Internal.pm
@@ -25,6 +25,8 @@ use MooseX::Types -declare => [
         Blog
         PerlMongers
         Tests
+        RTIssueStatus
+        GitHubIssueStatus
         BugSummary
         RiverSummary
         )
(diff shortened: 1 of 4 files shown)


mickeyn:
still having some issues with the tests... debugging

mickeyn:
@oalders do you know why [this line](https://github.com/metacpan/metacpan-api/blob/master/t/00_setup.t#L95) points to non-existing directory?
I'm missing the GH info for the tests to pass (type validation)

mickeyn:
@oalders fixed the issue exposed by the tests (made the 'rt' and 'bugs' attributes optional + fixed the expected structure in the tests)

lib/MetaCPAN/Types/Internal.pm
```diff
@@ -93,16 +95,27 @@ coerce Profile, from HashRef,
...
     (
         map { $_ => Optional [Int] }
-            qw(new open stalled patched resolved rejected active closed)
+            qw( active closed new open patched rejected resolved stalled )
     ),
-    type   => Str,
     source => Str
     ];
 
+subtype BugsGithub,
```
oalders: I like the custom types.  Could we rename them to `GitHubBugs` and `RTBugs`?  That would be in keeping with types like `PositiveNum` where the adjective precedes the noun.  I think that reads more like a natural language.  (Also, note the casing on `GitHub`).

oalders: On second thought, these aren't all bugs.  GitHub or RT issues might be wishlist items, so calling everything a bug might be misleading.

oalders: So, how about `GitHubIssueStatus` and `RTIssueStatus`?  Singular rather than plural would be in keeping with the other types.

PR #517, opened 2016-09-01
fix download_url version filtering notation parsing (!=, <=, etc.)

### LINKED ISSUE #515: v1 download_url endpoint does not support `!=` notation
I'm working on moving over Menlo::Index::MetaCPAN to v1 API.
Then I found that v1 download_url endpoint did not support `!=` notition.

Eg:
https://fastapi.metacpan.org/v1/download_url/Moose?version=!%3D+2.1805
(here `!%3D+2.1805` means `!= 2.1805`)
expect: 200 OK, Moose with version 2.1804
actual: 500 Internal Server Error

cf https://gist.github.com/skaji/f08b71f44866c0951fec460b3e2730df

mickeyn:
fixes #515 + range notation

### DIFF
--- a/lib/MetaCPAN/Document/File/Set.pm
+++ b/lib/MetaCPAN/Document/File/Set.pm
@@ -199,6 +199,8 @@ sub find_download_url {
         : !$explicit_version ? { term => { maturity => 'released' } }
         :                      ();
 
+    my $version_filters = $self->_version_filters($version);
+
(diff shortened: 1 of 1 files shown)


oalders:
Looks good.  Should we add a regression test to this?

mickeyn:
we probably should :)

skaji:
😊

PR #521, opened 2016-09-08
Leo/fastly purge and headers

ranguard:


cpanfile
```diff
@@ -103,6 +105,7 @@ requires 'MooseX::ClassAttribute';
 requires 'MooseX::Getopt';
 requires 'MooseX::Getopt::Dashes';
 requires 'MooseX::Getopt::OptionTypeMap';
+requires 'MooseX::Fastly::Role';
```
oalders: Should we have a version requirement here?

lib/MetaCPAN/Role/Script.pm
```diff
@@ -10,9 +9,13 @@ use Log::Contextual qw( :log :dlog );
 use MetaCPAN::Model;
 use MetaCPAN::Types qw(:all);
 use MetaCPAN::Queue ();
-use Moose::Role;
+
 use Carp ();
 
+with 'MetaCPAN::Role::HasConfig';
+with 'MetaCPAN::Role::Fastly';
+with 'MetaCPAN::Role::Logger';
```
oalders: I think we want to prohibit multiple `with`s, unless there's a compelling reason. So, something like:

```perl
with( 'MetaCPAN::Role::Fastly', 'MetaCPAN::Role::HasConfig', 'MetaCPAN::Role::Logger',);
```

oalders: @ranguard https://metacpan.org/pod/Perl::Critic::Policy::Moose::ProhibitMultipleWiths

lib/MetaCPAN/Script/Release.pm
```diff
@@ -184,7 +184,8 @@ sub run {
     $self->index->refresh unless $self->queue;
 
     # Call Fastly to purge
-    $self->cdn_purge_cpan_distnameinfos( \@module_to_purge_dists );
+    $self->purge_cpan_distnameinfos( \@module_to_purge_dists );
+
```
oalders: Extra newline.

lib/MetaCPAN/Server.pm
```diff
@@ -3,6 +3,9 @@ package MetaCPAN::Server;
 use Moose;
 
 ## no critic (Modules::RequireEndWithOne)
+use Catalyst qw/
+    +MetaCPAN::Role::Fastly::Catalyst
+    /;
```
oalders: No need to fix now, but it would be nice if we were consistent with quoting here.

```perl
use Catalyst qw( +MetaCPAN::Role::Fastly::Catalyst );
```

lib/MetaCPAN/Server/Controller/Author.pm
```diff
@@ -24,8 +24,13 @@ __PACKAGE__->config(
     }
 );
 
+# https://fastapi.metacpan.org/v1/author/LLAP
```
oalders: 👍

lib/MetaCPAN/Server/Controller/Changes.pm
```diff
@@ -19,9 +19,13 @@ has '+type' => ( default => 'file' );
 sub index : Chained('/') : PathPart('changes') : CaptureArgs(0) {
 }
 
+# https://fastapi.metacpan.org/v1/changes/LLAP/CatalystX-Fastly-Role-Response-0.04
```
oalders: 👍

lib/MetaCPAN/Server/Controller/Diff.pm
```diff
@@ -19,6 +19,10 @@ sub index : Chained('/') : PathPart('diff') : CaptureArgs(0) {
 sub diff_releases : Chained('index') : PathPart('release') : Args(4) {
     my ( $self, $c, @path ) = @_;
 
+    $c->add_author_key( $path[0] );
+    $c->add_author_key( $path[2] );
```
oalders: I guess there's no harm if the author names are the same?

ranguard: No harm, but I'll add a check in anycase

Issue #522, opened 2016-09-08
V1 API does not support 'source' queries


TiMBuS:
cpanm uses 'source' queries to avoid POST requests (for whatever reason). 
See the *Note* section here for an example: https://github.com/metacpan/metacpan-api/blob/master/docs/API-docs.md#user-content-downstream-dependencies

Running the linked example curl request against fastapi.metacpan.org/v1/ throws this error:
```
{
   "message" : "[Param] ** Unknown param (source) in (search) request. See docs at: http://www.elastic.co/guide/en/elasticsearch/reference/current/search-search.html, called from sub Search::Elasticsearch::Role::Client::Direct::__ANON__ at /home/metacpan/metacpan-api/lib/MetaCPAN/Server/Controller.pm line 110."
}
```
Which sadly breaks cpanm/menlo.

skaji:
I think cpanm will use download_url endpoint.

TiMBuS:
From what I can see, Menlo tries the following query first:

```json
localhost:5000/file/_search?source={
  "fields": [
    "date",
    "release",
    "author",
    "module",
    "status"
  ],
  "query": {
    "filtered": {
      "query": {
        "nested": {
          "path": "module",
          "score_mode": "max",
          "query": {
            "custom_score": {
              "query": {
                "constant_score": {
                  "filter": {
                    "and": [
                      {
                        "term": {
                          "module.authorized": true
                        }
                      },
                      {
                        "term": {
                          "module.indexed": true
                        }
                      },
                      {
                        "term": {
                          "module.name": "TestModule"
                        }
                      }
                    ]
                  }
                }
              },
              "metacpan_script": "score_version_numified"
            }
          }
        }
      },
      "filter": {
        "and": [
          {
            "term": {
              "maturity": "released"
            }
          }
        ]
      }
    }
  }
}
```

It will then query `release/_search?source={{release & author}}` and finally fish out the download_url. This would all still work if 'source' was working.

mickeyn:
@TiMBuS maybe I misunderstood, but you can't expect ES queries for v0 (ES0.20) to simply work on v1 (ES2.3)

ranguard:
@mickeyn @TiMBuS is playing with our interface.

@TiMBuS please use our `legacy/v0` branches (for puppet web and api) if you do not want to play with the new version of ES, but we no longer support this. At the same time not all the tools have caught up with our v1 (what is currently on master), which we haven't deployed to production yet as it's still under heavy development.

[...]

PR #523, opened 2016-09-09
fix 'source' parameter for request body

mickeyn:
for #522 

not sure if that's a complete solution, but it works for me.

### DIFF
--- a/lib/MetaCPAN/Server/Controller.pm
+++ b/lib/MetaCPAN/Server/Controller.pm
@@ -112,7 +112,7 @@ sub search : Path('_search') : ActionClass('Deserialize') {
                 {
                     index => $c->model('CPAN')->index,
                     type  => $self->type,
-                    body  => $c->req->data,
+                    body  => $c->req->data || delete $params->{source},
                     %$params,
                 }
             )


lib/MetaCPAN/Server/Controller.pm
```diff
@@ -112,7 +112,7 @@ sub search : Path('_search') : ActionClass('Deserialize') {
                 {
                     index => $c->model('CPAN')->index,
                     type  => $self->type,
-                    body  => $c->req->data,
+                    body  => $c->req->data || delete $params->{source},
```
oalders: Wow, that seems easy.  :)

oalders:
Is this still a WIP?

mickeyn:
I just wanted it tested to know it solves the original issue reported, but I don't mind having it merged.

mickeyn:
ok, since it does solve the problem of passing the request body through 'source' over GET - we should merge it.
original issue also have ES syntax problems that need to fix fixed on the client end separately.

oalders:
Ok.  Will merge, but I'll add an issue to create a test for this.

Issue #524, opened 2016-09-10
Add test for 'source' parameter as used by cpanm


oalders:
Add a test for #523

mickeyn:
test was added

PR #525, opened 2016-09-10
add test for the URL parameters: test 'source'

mickeyn:
#524 

created a generic test file for URL parameters, in case we have more to add tests for.


oalders:
Thanks!

PR #529, opened 2016-09-26
queue 'latest' jobs from 'release' (if queue is used)

mickeyn:
in the case we are using `--queue` & `--latest` in sctipt/release - queue the 'latest' jobs in lower priority instead of executing script from within a script


oalders:
Thank you, sir!

PR #530, opened 2016-09-27
[WIP] created new endpoints to remove web ES awareness

mickeyn:
as discussed before, i'd like to start moving some ES queries done in metacpan-web to the API as new endpoints.

the idea is to hide the storage behind the API and allow consolidation of functionality between all API using tools (web, client, etc.)

this is just a first step (out of many) so the PR is mostly for the welcomed discussion.

### DIFF
--- a/lib/MetaCPAN/Server/Controller/Author.pm
+++ b/lib/MetaCPAN/Server/Controller/Author.pm
@@ -60,11 +60,51 @@ sub get : Path('') : Args(1) {
         ['The requested field(s) could not be found'] );
 }
 
+# endpoint: /author/search?key=<key>[&fields=<csv_fields>][&sort=<csv_sort>][&size=N]
+sub search : Path('search') : Args(0) {
+    my ( $self, $c ) = @_;
+    my $key    = $c->req->parameters->{key};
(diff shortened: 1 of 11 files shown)


lib/MetaCPAN/Server/Controller/Author.pm
```diff
@@ -54,4 +55,11 @@ sub get : Path('') : Args(1) {
         ['The requested field(s) could not be found'] );
 }
 
+# endpoint: /author/by_user?users=<csv_user_ids>[&fields=<csv_fields>][&sort=<csv_sort>][&size=N]
+sub by_user : Path('by_user') : Args(0) {
+    my ( $self, $c ) = @_;
+    my @users = split /,/ => $c->req->parameters->{users};
+    $self->es_query_by_key( $c, 'user', \@users );
+}
```
oalders: Should this be a role? It looks to be identical in two of these packages.

mickeyn: slowly getting there :)

mickeyn:
most of this is currently happening in small chunks, closing the PR

PR #532, opened 2016-11-01
Always use the index alias instead of the name

mickeyn:
We should always use the alias of the index ('cpan') rather than the actual index name ('cpan_v1') to allow hot-swapping of newly created indices (in some of the mapping changing cases like #516)


mickeyn:
for reference: https://www.elastic.co/blog/changing-mapping-with-zero-downtime

mickeyn:
tests are failing... will investigate

PR #535, opened 2016-11-11
script role: ua attribute (lazy) + proxy attribute + cleanup

mickeyn:
made shared `ua` and `proxy` attributes available for all scripts.
cleaned up the setting from 2 scripts.


oalders:
👍

Issue #534, opened 2016-11-07
'module' entry reports wrong module


xsawyerx:
The way the **module** endpoint works is wrong, probably because of the underlying structure treating modules as files. Instead of giving you the definite and authoritative module, it gives you a result of a module it finds[1].

Example:
https://v1.metacpan.org/module/warnings

This leads to `Apache::Test` instead of `warnings`.

You can see it here as well:
```
$ perl -MMetaCPAN::Client -E'say MetaCPAN::Client->new->module("warnings")->distribution'
Apache-Test
```

[1] I don't know if it's the first it finds, the first it indexed, or what other metric it uses to pick one out of many.

ranguard:
Fixed

haarg:
This is still an issue for https://metacpan.org/pod/overload for similar reasons.  The scoring used in the query doesn't seem reliable now that we are using sharding.

PR #546, opened 2016-11-17
Fixes Issue #534

reyjrar:
The sort being performed was ignoring the search score.  This short
circuited the 'should' section of the query completely.  This patch
sets the _score as the primary sort criteria, and falls back on the
existing criteria to find the latest, most up-to-date file.


mickeyn:
related to #534

oalders:
Thanks!

mickeyn:
problem still seem to exist

PR #547, opened 2016-11-18
[WIP] set the alias_for value for the index hot-swap

mickeyn:
merge AFTER the alias hotswap


mickeyn:
will break tests, we'll just push after the hotswap

PR #556, opened 2016-11-18
Find module query revamp.

reyjrar:
Drawing on the work from find_download_url, simplify the query to return
modules with valid POD entries.  Return documents where either
the documentation and pod contain the module name, *OR* the
module.associated_pod is set with module.name matching the request path.

Using 'minimum_should_match=1' to ensure that a valid .pm/POD is
selected.


reyjrar:
This should solve issue #534 for reals.

PR #570, opened 2016-11-20
/search/autocomplete: Collapse single-valued arrays in fields

tsibley:
To be consistent with API v0 and other v1 endpoints.  No need to leak
more crazy from Elasticsearch than necessary.


tsibley:
The one Travis fail is due to untidy other files that have since been corrected in master.

PR #579, opened 2016-11-22
Add web and simple search endpoints

jberger:
The web search is the true gain here, the simple search avoids most of the input and output aggregation and munging and is therefore easier to reason about for say search optimization.


lib/MetaCPAN/Model/Search.pm
```diff
@@ -0,0 +1,484 @@
+package MetaCPAN::Model::Search;
+
+use Moose;
```
oalders: Any reason not to use `MetaCPAN::Moose` here?  That gives you `MooseX::StrictConstructor` for free.

jberger: The only reason I can think of is that I didn't know about it (also, I copied from MetaCPAN::Model::Release which does't use it).

Issue #580, opened 2016-11-22
/v1/module handles fields parameter in a way that produces single-value arrays


tsibley:
Compare:

```
$ curl -fsS https://api.metacpan.org/v0/module/Pod::Cpandoc?fields=distribution
{
   "distribution" : "Pod-Cpandoc"
}

$ curl -fsS https://fastapi.metacpan.org/v1/module/Pod::Cpandoc?fields=distribution
{
   "distribution" : [
      "Pod-Cpandoc"
   ]
}

$ curl -fsS https://fastapi.metacpan.org/v1/module/Pod::Cpandoc
{
   …,
   "distribution" : "Pod-Cpandoc",
   …
}
```

I'll take a look at this in the next couple days if no one else does.

ranguard:
I know the 'utility' of flattening arrays is nice, but if it leads to our code having lots of these we should seriously consider just keeping them as arrays all  the time

ranguard:
s/keeping them/keeping results/

tsibley:
Well, it's a huge PITA for consumers which otherwise would just be able to change endpoints and be done with it.

tsibley:
For this particular endpoint, we obviously take pains to collapse single-valued arrays, but miss the case when the `fields` parameter is specified or something.  (I haven't looked at the code yet.)

tsibley:
We can also refactor the controller/model code such that there's only a few places where the output gets filtered through our munger.

tsibley:
Fixed in #581.

PR #581, opened 2016-11-23
Provide backwards compatibility for clients which request only specific fields

tsibley:
We claim in our docs¹ that our convenience endpoints revert the newer
Elasticsearch behaviour of wrapping single values in arrays.  This makes
that so in more places, specifically the places where the client has
requested a limited subset of fields.

While this may seem like a pain for us and lead to uglier code, think
about it this way: either we do it for our API one time, or every client
that runs into the issue has to do it themselves many times over.

A better solution may exist which centralizes application of
single_valued_arrayref_to_scalar(), but this works for now.

See also GH#580 for an example case.²

¹ https://github.com/metacpan/metacpan-examples/blob/master/README.md#upgrading-from-v0
² https://github.com/metacpan/metacpan-api/issues/580

### DIFF
--- a/lib/MetaCPAN/Server/Controller.pm
+++ b/lib/MetaCPAN/Server/Controller.pm
@@ -82,7 +82,8 @@ sub get : Path('') : Args(1) {
     if ( !defined $file ) {
         $c->detach( '/not_found', ['Not found'] );
     }
-    $c->stash( $file->{_source} || $file->{fields} )
+    $c->stash( $file->{_source}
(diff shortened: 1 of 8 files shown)


oalders:
Looks good.  Just has some failing tests.

tsibley:
Fixed the tests, which did uncover a subtle bug in `single_valued_arrayref_as_scalar` that I also fixed!

tsibley:
…and passing now.

PR #590, opened 2016-11-29
Notes on digging into the authorized and indexed flags

ranguard:
@tsibley - can we merge this :) ?


tsibley:
Yes, can be merged!

Issue #470, opened 2016-05-04
[mi/es2] investigate use of suggester


mickeyn:
see https://www.elastic.co/guide/en/elasticsearch/reference/current/search-suggesters-completion.html
as an optional alternative to our current `autocomplete` query

jberger:
Implemented via https://github.com/metacpan/metacpan-api/pull/764 and https://github.com/metacpan/metacpan-web/pull/1992

jberger:
And now note that we are changing the structure to include more information: https://github.com/metacpan/metacpan-api/pull/768

PR #595, opened 2016-12-01
[WIP] mapping deploy simplified

mickeyn:
so, 1st commit just changes the way we currently do (rarely) a rebuild of the index and mappings for all the types.
this commit removes the use of the `ElasticSearchX::Model` functionality to achieve this (so ignores the attributes' meta black magic) and adds simple JSON string methods to reflect the mappings for the types (to be fully ES compatible, easier to read and update)

the 2nd commit changes what I think is wrong with the current mapping - the module.name and documentation were marked as 'index => not_analyzed' which is probably due to wrong logic I introduced into the module when fighting all the fires of adapting it to ES 2.x.
I am sure there are other fields that probably should be changed as well (they are assigned with analyzers but marked as 'not_analyzed', but we can do those one later (even though it does require cloning a new index)

please review the changes first, to make sure the queries are not being completely broken here (tests work)

to deploy this we will need an update to the WEB repo (a parallel branch with the same name, though a small fix there) and replacing the index with a new clone with the updated mapping (I will prepare this one when we're ready)

### DIFF
--- a/lib/MetaCPAN/Document/File.pm
+++ b/lib/MetaCPAN/Document/File.pm
@@ -363,6 +363,33 @@ sub _build_documentation {
     return undef;
 }
 
+=head2 suggest
+
+Autocomplete suggester data.
+
+=cut
+
+has suggest => (
+    is      => 'ro',
+    lazy    => 1,
+    builder => '_build_suggest',
+);
+
+sub _build_suggest {
+    my $self = shift;
+    my $doc  = $self->documentation;
+    return unless $doc;
+
+    my $weight = 1000 - length($doc);
+    $weight = 0 if $weight < 0;
+
+    return {
+        input   => [$doc],
+        payload => { doc_name => $doc },
+        weight  => $weight
+    };
+}
+
 =head2 indexed
 
 B<Default 0>
--- a/lib/MetaCPAN/Document/File/Set.pm
+++ b/lib/MetaCPAN/Document/File/Set.pm
@@ -31,7 +31,7 @@ sub find {
                                     filter => {
                                         and => [
                                             {
-                                                term => {
+                                                match => {
                                                     "module.name" => $module
                                                 }
                                             },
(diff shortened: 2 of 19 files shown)


mickeyn:
not sure why this test fails... it passes on my VM

ranguard:
The theory of this looks good... :)

mickeyn:
ok, now the autocomplete test fails on my VM and I can actually debug it

mickeyn:
Following some hacking with @ranguard, a completion suggester was added to address the autocomplete issue.
It makes sense to add it here, as ElasticSearchX::Model will require changes to support adding it there, and we are in the process of moving away from it.
The current `suggester` implementation is probably not optimal and the last commit introduces an issue with a different test that needs to be investigated, but overall I think this is going in the right direction.

mickeyn:
yay. all tests were fixed.

mickeyn:
this will solve #470

lib/MetaCPAN/Document/File/Set.pm
```diff
@@ -442,37 +444,57 @@ sub autocomplete {
     my $query = join( q{ }, @terms );
     return $self unless $query;
 
-    return $self->search_type('dfs_query_then_fetch')->query(
+    my $suggestions = $self->search_type('dfs_query_then_fetch')->es->search(
```
ranguard: is there `es->suggest`?

lib/MetaCPAN/Document/File.pm
```diff
@@ -363,6 +363,32 @@ sub _build_documentation {
...
+    lazy    => 1,
+    builder => '_build_suggest',
+);
+
+sub _build_suggest {
+    my $self = shift;
+    my $doc  = $self->documentation;
+    return unless $doc;
+
+    my $weight = 1000 - length($doc);
```
mickeyn: adding

mickeyn:
@reyjrar, would like to have your review of this PR.
I bet there's more we can improve on the mapping/queries side.

mickeyn:
closing this in favor of deploying the changes in steps to allow proper migration with `metacpan-web`.
see #599

PR #597, opened 2016-12-06
default type copy to monthly slices (unless query is provided)

mickeyn:
make `type` copying a bit more automatic.


mickeyn:
seems ready to merge.
@ranguard we'll need this to prepare a new clone for #595

PR #598, opened 2016-12-08
script/mapping: improved index cloning

mickeyn:


lib/MetaCPAN/Script/Mapping.pm
```diff
@@ -226,16 +221,18 @@ sub index_create {
     if @patch_types;
 }
 
-sub copy_index {
-    my $self = shift;
+sub type_copy {
```
oalders: Any reason why this is `type_copy` rather than `copy_type`?

mickeyn: `copy_type` is taken by the attribute name :)

mickeyn: This is fixed in #599 as part of a bigger renaming work for attributes/methods and requires a newer version of `MooseX::Getopt`.
I suggest we leave this one as is for this merge, and let the other PR merge its changes later.

PR #599, opened 2016-12-11
mapping simplified

mickeyn:
As a first step, this will not make changes to the current mapping description of `file.documentation` or `file.module.name`.
It does add the `file.suggest` field (though this is not immediately affecting anything without rebuilding of the index)

The plan is for this PR is to:
- merge these changes (no impact)
- add the suggest mapping to current index (not used)
- back-fill the suggest values using the new script
- new release indexing will automatically add the values per the `Document::File` + `Script::Release` change. 

Once that's done, and suggest is we will be able to test `metacpan-web` against a live version of the new autocomplete before phasing out the old one.

### DIFF
--- a/lib/MetaCPAN/Script/Mapping.pm
+++ b/lib/MetaCPAN/Script/Mapping.pm
@@ -10,13 +10,13 @@ use Cpanel::JSON::XS qw( decode_json );
 use DateTime;
 
 use MetaCPAN::Script::Mapping::DeployStatement;
-use MetaCPAN::Script::Mapping::Cpan::Author;
-use MetaCPAN::Script::Mapping::Cpan::Distribution;
-use MetaCPAN::Script::Mapping::Cpan::Favorite;
-use MetaCPAN::Script::Mapping::Cpan::File;
-use MetaCPAN::Script::Mapping::Cpan::Mirror;
-use MetaCPAN::Script::Mapping::Cpan::Rating;
-use MetaCPAN::Script::Mapping::Cpan::Release;
+use MetaCPAN::Script::Mapping::CPAN::Author;
+use MetaCPAN::Script::Mapping::CPAN::Distribution;
+use MetaCPAN::Script::Mapping::CPAN::Favorite;
+use MetaCPAN::Script::Mapping::CPAN::File;
+use MetaCPAN::Script::Mapping::CPAN::Mirror;
+use MetaCPAN::Script::Mapping::CPAN::Rating;
+use MetaCPAN::Script::Mapping::CPAN::Release;
 use MetaCPAN::Script::Mapping::User::Account;
 use MetaCPAN::Script::Mapping::User::Identity;
 use MetaCPAN::Script::Mapping::User::Session;
@@ -391,19 +391,19 @@ sub deploy_mapping {
     my %mappings = (
         $idx_cpan => {
             author =>
-                decode_json(MetaCPAN::Script::Mapping::Cpan::Author::mapping),
+                decode_json(MetaCPAN::Script::Mapping::CPAN::Author::mapping),
             distribution =>
-                decode_json( MetaCPAN::Script::Mapping::Cpan::Distribution::mapping
+                decode_json( MetaCPAN::Script::Mapping::CPAN::Distribution::mapping
                 ),
             favorite =>
-                decode_json( MetaCPAN::Script::Mapping::Cpan::Favorite::mapping
+                decode_json( MetaCPAN::Script::Mapping::CPAN::Favorite::mapping
                 ),
             file =>
-                decode_json(MetaCPAN::Script::Mapping::Cpan::File::mapping),
+                decode_json(MetaCPAN::Script::Mapping::CPAN::File::mapping),
             rating =>
-                decode_json(MetaCPAN::Script::Mapping::Cpan::Rating::mapping),
+                decode_json(MetaCPAN::Script::Mapping::CPAN::Rating::mapping),
             release =>
-                decode_json( MetaCPAN::Script::Mapping::Cpan::Release::mapping
+                decode_json( MetaCPAN::Script::Mapping::CPAN::Release::mapping
                 ),
         },
         $idx_user => {
--- a/lib/MetaCPAN/Script/Mapping.pm
+++ b/lib/MetaCPAN/Script/Mapping.pm
@@ -9,6 +9,18 @@ use MetaCPAN::Types qw( Bool Str );
 use Cpanel::JSON::XS qw( decode_json );
 use DateTime;
 
+use MetaCPAN::Script::Mapping::DeployStatement;
+use MetaCPAN::Script::Mapping::Cpan::Author;
+use MetaCPAN::Script::Mapping::Cpan::Distribution;
+use MetaCPAN::Script::Mapping::Cpan::Favorite;
+use MetaCPAN::Script::Mapping::Cpan::File;
+use MetaCPAN::Script::Mapping::Cpan::Mirror;
+use MetaCPAN::Script::Mapping::Cpan::Rating;
+use MetaCPAN::Script::Mapping::Cpan::Release;
+use MetaCPAN::Script::Mapping::User::Account;
+use MetaCPAN::Script::Mapping::User::Identity;
+use MetaCPAN::Script::Mapping::User::Session;
+
 use constant {
     EXPECTED     => 1,
     NOT_EXPECTED => 0,
@@ -109,7 +121,7 @@ sub run {
     $self->type_copy      if $self->copy_to_index;
     $self->type_empty     if $self->delete_from_type;
     $self->types_list     if $self->list_types;
-    $self->delete_mapping if $self->delete;
+    $self->deploy_mapping if $self->delete;
 }
 
 sub _check_index_exists {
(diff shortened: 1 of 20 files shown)


mickeyn:
added `search/autocomplete/_get` (to later replace `sub get`) endpoint for testing `metacpan-web`

mickeyn:
We should probably add the mapping to the current index before merging. 
To do so, we can run this line:
```
./bin/run ./bin/metacpan mapping --update_index cpan_v1_01 --patch_mapping '{"file":{"properties":{"suggest":{"type":"completion","analyzer":"simple","payloads":true,"preserve_separators":true,"preserve_position_increments":true,"max_input_length":50}}}}'
```

lib/MetaCPAN/Document/File.pm
```diff
@@ -363,6 +363,33 @@ sub _build_documentation {
...
 
+=head2 suggest
+
+Autocomplete suggester data.
+
+=cut
+
+has suggest => (
+    is      => 'ro',
+    lazy    => 1,
```
oalders: `isa => Maybe[HashRef]`?

lib/MetaCPAN/Document/File.pm
```diff
@@ -363,6 +363,33 @@ sub _build_documentation {
...
+    my $doc  = $self->documentation;
+    return unless $doc;
+
+    my $weight = 1000 - length($doc);
+    $weight = 0 if $weight < 0;
+
+    return {
+        input   => [$doc],
+        payload => { doc_name => $doc },
+        weight  => $weight
```
oalders: Missing trailing comma.

lib/MetaCPAN/Document/File/Set.pm
```diff
@@ -475,5 +477,63 @@ sub autocomplete {
     )->sort( [ '_score', 'documentation' ] );
 }
 
+# this method will replace 'sub autocomplete' after the
+# mapping + data is fully deployed.
+# -- Mickey
+sub autocomplete_new {
```
oalders: Can we find a name that doesn't include a `_new` suffix?  I see these at $work and they're baffling to everybody who wasn't around when they were created.

lib/MetaCPAN/Script/Mapping.pm
```diff
@@ -9,6 +9,18 @@ use MetaCPAN::Types qw( Bool Str );
...
+use MetaCPAN::Script::Mapping::Cpan::Author;
+use MetaCPAN::Script::Mapping::Cpan::Distribution;
+use MetaCPAN::Script::Mapping::Cpan::Favorite;
+use MetaCPAN::Script::Mapping::Cpan::File;
+use MetaCPAN::Script::Mapping::Cpan::Mirror;
+use MetaCPAN::Script::Mapping::Cpan::Rating;
+use MetaCPAN::Script::Mapping::Cpan::Release;
+use MetaCPAN::Script::Mapping::User::Account;
+use MetaCPAN::Script::Mapping::User::Identity;
+use MetaCPAN::Script::Mapping::User::Session;
```
oalders: Could we alpha-sort all of these?

lib/MetaCPAN/Script/Mapping.pm
```diff
@@ -134,6 +146,11 @@ sub index_delete {
     $self->_check_index_exists( $name, EXPECTED );
     $self->are_you_sure("Index $name will be deleted !!!");
 
+    $self->_index_delete($name);
+}
+
+sub _index_delete {
```
oalders: Any reason not to call this `_delete_index`?

lib/MetaCPAN/Script/Mapping.pm
```diff
@@ -346,13 +363,85 @@ sub types_list {
     print "$_\n" for sort keys %{ $self->index->types };
 }
 
-sub delete_mapping {
-    my $self = shift;
+sub deploy_mapping {
+    my $self     = shift;
+    my $es       = $self->es;
+    my $idx_cpan = 'cpan_v1_01';
+    my $idx_user = 'user';
```
oalders: Any reason not to call these `$cpan_index` and `$user_index`?

lib/MetaCPAN/Script/Mapping.pm
```diff
@@ -346,13 +363,85 @@ sub types_list {
...
+    # delete cpan (aliased) + user indices
+
+    $self->_index_delete($idx_user)
+        if $es->indices->exists( index => $idx_user );
+    $self->_index_delete($idx_cpan)
+        if $es->indices->exists( index => $idx_cpan );
+
+    # create new indices
+
+    my $dep = decode_json MetaCPAN::Script::Mapping::DeployStatement::mapping;
```
oalders: `my $dep = decode_json( MetaCPAN::Script::Mapping::DeployStatement::mapping );`

lib/MetaCPAN/Script/Mapping/Cpan/Author.pm
```diff
@@ -0,0 +1,163 @@
+package MetaCPAN::Script::Mapping::Cpan::Author;
```
oalders: s/Cpan/CPAN/

lib/MetaCPAN/Script/Mapping/Cpan/Distribution.pm
```diff
@@ -0,0 +1,91 @@
+package MetaCPAN::Script::Mapping::Cpan::Distribution;
```
oalders: s/Cpan/CPAN/

lib/MetaCPAN/Script/Mapping/Cpan/Favorite.pm
```diff
@@ -0,0 +1,43 @@
+package MetaCPAN::Script::Mapping::Cpan::Favorite;
```
oalders: s/Cpan/CPAN/

lib/MetaCPAN/Script/Mapping/Cpan/File.pm
```diff
@@ -0,0 +1,300 @@
+package MetaCPAN::Script::Mapping::Cpan::File;
```
oalders: s/Cpan/CPAN/

lib/MetaCPAN/Script/Mapping/Cpan/Mirror.pm
```diff
@@ -0,0 +1,175 @@
+package MetaCPAN::Script::Mapping::Cpan::Mirror;
```
oalders: s/Cpan/CPAN/

lib/MetaCPAN/Script/Mapping/Cpan/Rating.pm
```diff
@@ -0,0 +1,64 @@
+package MetaCPAN::Script::Mapping::Cpan::Rating;
```
oalders: s/Cpan/CPAN/

lib/MetaCPAN/Script/Mapping/Cpan/Release.pm
```diff
@@ -0,0 +1,267 @@
+package MetaCPAN::Script::Mapping::Cpan::Release;
```
oalders: s/Cpan/CPAN/

lib/MetaCPAN/Script/Suggest.pm
```diff
@@ -0,0 +1,117 @@
+package MetaCPAN::Script::Suggest;
+
+use strict;
+use warnings;
+
+use Log::Contextual qw( :log );
+use Moose;
+use MetaCPAN::Types qw( Bool Int );
+use DateTime;
```
oalders: Could we alpha-sort these?

Also, it's nice to be explicit about not importing anything.  ie `use DateTime ()`.

lib/MetaCPAN/Script/Suggest.pm
```diff
@@ -0,0 +1,117 @@
...
+    default       => 0,
+    documentation => 'update all records',
+);
+
+sub run {
+    my $self = shift;
+
+    if ( $self->all ) {
+        my $dt = DateTime->new( year => 1994, month => 1 );
+        my $end_time = DateTime->now()->add( months => 1 );
```
oalders: `my $end_time = DateTime->now->add( months => 1 );`

lib/MetaCPAN/Script/Suggest.pm
```diff
@@ -0,0 +1,117 @@
...
+
+__END__
+
+=head1 SYNOPSIS
+
+ # bin/metacpan suggest
+
+=head1 DESCRIPTION
+
+After importing releases from cpan, this script will set the suggest
```
oalders: s/cpan/CPAN/

oalders:
I'm not a huge fan of having raw JSON in the mapping classes, but if you think that's best, I'll go along with it.  It just means we now have to worry about trailing commas inside our Perl classes.

lib/MetaCPAN/Script/Mapping.pm
mickeyn: they are alpha-sorted, what you see here is because of the directory (`Cpan`/`User`).

lib/MetaCPAN/Script/Mapping.pm
mickeyn: the "proper" names were taken by the attributes. i'll come up with some postfix to improve the methods names, i didn't like it either.

mickeyn:
I wasn't happy about adding the raw JSON, but I believe we are most likely to update these values through copy-pasting from the ES _mapping output - so the source is raw JSON.

Also, I wanted it to be very clear when reading the mapping what we have in the ES index and with the slight differences (like `true` & `false` values) I rather have a more ES compatible representation.

We can always change it later if it turns out to be a problem.

mickeyn:
@oalders I pushed a change to address all your comments. hope it's better.
I hope the attribute/method name workaround is acceptable.

oalders:
Thanks, @mickeyn! No blockers on my end.

PR #600, opened 2016-12-13
Script for ES Snapshots

ranguard:
You can:

- setup
- snap
- list
- restore

Built so that it's easy to put quite complex configs into cron

### DIFF
--- a/lib/MetaCPAN/Script/Snapshot.pm
+++ b/lib/MetaCPAN/Script/Snapshot.pm
@@ -3,16 +3,14 @@ package MetaCPAN::Script::Snapshot;
 use strict;
 use warnings;
 
+use Cpanel::JSON::XS qw(encode_json decode_json);
+use DateTime ();
+use DDP qw(np);
+use HTTP::Tiny ();
 use Log::Contextual qw( :log );
-
-use MetaCPAN::Types qw( Bool Int Str File );
+use MetaCPAN::Types qw( Bool Int Str File ArrayRef );
 use Moose;
-use DateTime;
-use Try::Tiny;
-use Sys::Hostname;
-use HTTP::Tiny;
-use Cpanel::JSON::XS;
-use DDP;
+use Sys::Hostname qw(hostname);
 
 with 'MetaCPAN::Role::Script', 'MooseX::Getopt::Dashes';
 
@@ -28,50 +26,54 @@ my $repository_name = 'our_backups';
 has setup => (
     is            => 'ro',
     isa           => Bool,
+    default       => 0,
     documentation => 'Setup the connection with ES',
 );
 
 has snap => (
     is            => 'ro',
     isa           => Bool,
+    default       => 0,
     documentation => 'Perform a snapshot',
 );
 
 has list => (
     is            => 'ro',
     isa           => Bool,
+    default       => 0,
     documentation => 'List saved snapshots',
 );
 
 has restore => (
     is            => 'ro',
     isa           => Bool,
+    default       => 0,
     documentation => 'Perform a restore',
 );
 
 ## Options
 has snap_stub => (
     is  => 'ro',
-    isa => 'Str',
+    isa => Str,
     documentation =>
         'Stub of snapshot name ( e.g full, user etc ), used with dateformat to create the actual name in S3',
 );
 
 has date_format => (
     is            => 'ro',
-    isa           => 'Str',
+    isa           => Str,
     documentation => 'strftime format to add to snapshot name (eg %Y-%m-%d)',
 );
 
 has snap_name => (
     is            => 'ro',
-    isa           => 'Str',
+    isa           => Str,
     documentation => 'Full name of snapshot to restore',
 );
 
 has host => (
     is            => 'ro',
-    isa           => 'Str',
+    isa           => Str,
     default       => 'http://localhost:9200',
     documentation => 'ES host, defaults to: http://localhost:9200',
 );
@@ -79,7 +81,7 @@ has host => (
 # Note: can take wild cards https://www.elastic.co/guide/en/elasticsearch/reference/2.4/multi-index.html
 has indices => (
     is      => 'ro',
-    isa     => 'ArrayRef',
+    isa     => ArrayRef,
     default => sub { ['*'] },
     documentation =>
         'Which indices to snapshot, defaults to "*" (all), can take wild cards - "*v100*"',
@@ -140,9 +142,9 @@ sub run_snapshot {
 
     my $indices = join ',', @{ $self->indices };
     my $data = {
-        "indices"              => $indices,
         "ignore_unavailable"   => 0,
-        "include_global_state" => 1
+        "include_global_state" => 1,
+        "indices"              => $indices,
     };
 
     log_debug { 'snapping: ' . $snap_name };
@@ -152,7 +154,6 @@ sub run_snapshot {
 
     my $response = $self->_request( 'put', $path, $data );
     return $response;
-
 }
 
 sub run_list_snaps {
@@ -169,7 +170,6 @@ sub run_list_snaps {
     }
 
     return $response;
-
 }
 
 sub run_restore {
@@ -179,12 +179,12 @@ sub run_restore {
 
     $self->are_you_sure('Restoring... will rename indices to restored_XX');
 
-    # This is a safetly feature, we can always
+    # This is a safety feature, we can always
     # create aliases to point to them if required
     # just make sure there is enough disk space
     my $data = {
         "rename_pattern"     => '(.+)',
-        "rename_replacement" => 'restored_$1'
+        "rename_replacement" => 'restored_$1',
     };
 
     my $path = "${repository_name}/${snap_name}/_restore";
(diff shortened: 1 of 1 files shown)


lib/MetaCPAN/Script/Snapshot.pm
```diff
@@ -0,0 +1,289 @@
...
+}
+
+__PACKAGE__->meta->make_immutable;
+1;
+
+__END__
+
+=head1 NAME
+
+MetaCPAN::Script::Backup - Backup indices and types
```
mickeyn: you probably want to change the name here :)

lib/MetaCPAN/Script/Snapshot.pm
```diff
@@ -0,0 +1,289 @@
...
+use Log::Contextual qw( :log );
+
+use MetaCPAN::Types qw( Bool Int Str File );
+use Moose;
+use DateTime;
+use Try::Tiny;
+use Sys::Hostname;
+use HTTP::Tiny;
+use Cpanel::JSON::XS;
+use DDP;
```
oalders: Could we alpha-sort these and explicitly not import anything by default? ie `use DateTime ();`

lib/MetaCPAN/Script/Snapshot.pm
```diff
@@ -0,0 +1,289 @@
...
+# So we dont' break production
+my $bucket = "mc-${mode}-backups";
+
+my $repository_name = 'our_backups';
+
+## Modes
+has setup => (
+    is            => 'ro',
+    isa           => Bool,
+    documentation => 'Setup the connection with ES',
```
oalders: Could we set explicit defaults for the `Bool` attributes?  I'm assuming they default to `false`, but I just tried to look it up and the Moose docs weren't clear to me.

lib/MetaCPAN/Script/Snapshot.pm
```diff
@@ -0,0 +1,289 @@
...
+has restore => (
+    is            => 'ro',
+    isa           => Bool,
+    documentation => 'Perform a restore',
+);
+
+## Options
+has snap_stub => (
+    is  => 'ro',
+    isa => 'Str',
```
oalders: You've imported `Str`, so shouldn't be quoted.

lib/MetaCPAN/Script/Snapshot.pm
```diff
@@ -0,0 +1,289 @@
...
+    documentation =>
+        'Stub of snapshot name ( e.g full, user etc ), used with dateformat to create the actual name in S3',
+);
+
+has date_format => (
+    is            => 'ro',
+    isa           => 'Str',
```
oalders: You've imported `Str`, so shouldn't be quoted.

lib/MetaCPAN/Script/Snapshot.pm
```diff
@@ -0,0 +1,289 @@
...
+    documentation => 'strftime format to add to snapshot name (eg %Y-%m-%d)',
+);
+
+has snap_name => (
+    is            => 'ro',
+    isa           => 'Str',
```
oalders: You've imported `Str`, so shouldn't be quoted.

lib/MetaCPAN/Script/Snapshot.pm
```diff
@@ -0,0 +1,289 @@
...
+    documentation => 'Full name of snapshot to restore',
+);
+
+has host => (
+    is            => 'ro',
+    isa           => 'Str',
```
oalders: You've imported `Str`, so shouldn't be quoted.

lib/MetaCPAN/Script/Snapshot.pm
```diff
@@ -0,0 +1,289 @@
...
+    default       => 'http://localhost:9200',
+    documentation => 'ES host, defaults to: http://localhost:9200',
+);
+
+# Note: can take wild cards https://www.elastic.co/guide/en/elasticsearch/reference/2.4/multi-index.html
+has indices => (
+    is      => 'ro',
+    isa     => 'ArrayRef',
```
oalders: We should import `ArrayRef` and use that here.

lib/MetaCPAN/Script/Snapshot.pm
```diff
@@ -0,0 +1,289 @@
...
+    $self->date_format || die 'Missing date-format (e.g. %Y-%m-%d)';
+
+    my $date      = DateTime->now->strftime( $self->date_format );
+    my $snap_name = $self->snap_stub . '_' . $date;
+
+    my $indices = join ',', @{ $self->indices };
+    my $data = {
+        "indices"              => $indices,
+        "ignore_unavailable"   => 0,
+        "include_global_state" => 1
```
oalders: Keys don't need to be quoted.  Missing trailing comma.  Could we alpha sort keys as well?

lib/MetaCPAN/Script/Snapshot.pm
```diff
@@ -0,0 +1,289 @@
...
+    };
+
+    log_debug { 'snapping: ' . $snap_name };
+    log_debug { 'with indices: ' . $indices };
+
+    my $path = "${repository_name}/${snap_name}";
+
+    my $response = $self->_request( 'put', $path, $data );
+    return $response;
+
```
oalders: Extra newline. ;)

lib/MetaCPAN/Script/Snapshot.pm
```diff
@@ -0,0 +1,289 @@
...
+    my $data = eval { decode_json $response->{content} };
+
+    foreach my $snapshot ( @{ $data->{snapshots} || [] } ) {
+        log_info { $snapshot->{snapshot} }
+        log_debug { np($snapshot) }
+    }
+
+    return $response;
+
```
oalders: Extra newline.

lib/MetaCPAN/Script/Snapshot.pm
```diff
@@ -0,0 +1,289 @@
...
+}
+
+sub run_list_snaps {
+    my $self = shift;
+
+    my $path = "${repository_name}/_all";
+    my $response = $self->_request( 'get', $path, {} );
+
+    my $data = eval { decode_json $response->{content} };
```
oalders: We're ok with nothing happening if the JSON doesn't parse?

lib/MetaCPAN/Script/Snapshot.pm
```diff
@@ -0,0 +1,289 @@
...
+}
+
+sub run_restore {
+    my $self = shift;
+
+    my $snap_name = $self->snap_name;
+
+    $self->are_you_sure('Restoring... will rename indices to restored_XX');
+
+    # This is a safetly feature, we can always
```
oalders: *safety

lib/MetaCPAN/Script/Snapshot.pm
```diff
@@ -0,0 +1,289 @@
...
+        "type"     => "s3",
+        "settings" => {
+            "bucket"                 => $bucket,
+            "region"                 => "us-east",
+            "protocol"               => "https",
+            "access_key"             => $self->aws_key,
+            "secret_key"             => $self->aws_secret,
+            "server_side_encryption" => 1,
+            "storage_class"          => "standard",
+            "canned_acl"             => "private",
```
oalders: Sort keys?  No need to quote keys here.

lib/MetaCPAN/Script/Snapshot.pm
```diff
@@ -0,0 +1,289 @@
...
+    my $response = $self->_request( 'put', $path, $data );
+    return $response;
+}
+
+sub _request {
+    my ( $self, $method, $path, $data ) = @_;
+
+    my $url = $self->host . '/_snapshot/' . $path;
+
+    my $json = encode_json $data;
```
oalders: Should probably use brackets unless the function is built in.

lib/MetaCPAN/Script/Snapshot.pm
```diff
@@ -0,0 +1,289 @@
...
+
+        log_error { 'Problem requesting ' . $url };
+
+        my $resp_json = eval { decode_json $response->{content} };
+        if ( my $error = $@ ) {
+            log_error { 'Error msg: ' . $response->{content} }
+        }
+        else {
+            log_error { 'Error response: ' . np($resp_json) }
+        }
```
oalders: Any reason not to use `Try::Tiny` here?

lib/MetaCPAN/Script/Snapshot.pm
```diff
@@ -0,0 +1,289 @@
...
+}
+
+__PACKAGE__->meta->make_immutable;
+1;
+
+__END__
+
+=head1 NAME
+
+MetaCPAN::Script::Snapshot - Snapshot (and restore) ElasticSearch indices
```
oalders: s/ElasticSearch/Elasticsearch/

oalders:
Don't feel like you have to change everything, but it would be nice to make those changes to the Moose attributes and also use `Try::Tiny` with new code.

lib/MetaCPAN/Script/Snapshot.pm
ranguard: You missed one - there was another extra new line! :)

lib/MetaCPAN/Script/Snapshot.pm
ranguard: Yea, this is on 'list' which you'd only ever run manually in anycase - actual error responses from ES are handled

lib/MetaCPAN/Script/Snapshot.pm
ranguard: Doh - ok - so you found it

PR #601, opened 2016-12-13
[WIP] autocomplete suggester

mickeyn:
An amendment to #599 ... all the changes that were lost in squashing + rebasing.

The `suggest` field here doesn't work yet (therefore WIP), needs review.

### DIFF
--- a/lib/MetaCPAN/Document/File.pm
+++ b/lib/MetaCPAN/Document/File.pm
@@ -371,7 +371,7 @@ Autocomplete info for documentation.
 
 has suggest => (
     is      => 'ro',
-    isa     => Maybe [HashRef],
+#    isa     => Maybe [HashRef], # remarked: breaks the suggester
     lazy    => 1,
     builder => '_build_suggest',
 );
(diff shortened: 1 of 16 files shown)


lib/MetaCPAN/Document/File.pm
```diff
@@ -363,6 +363,34 @@ sub _build_documentation {
...
 
+=head2 suggest
+
+Autocomplete info for documentation.
+
+=cut
+
+has suggest => (
+    is      => 'ro',
+    isa     => Maybe [HashRef],
```
ranguard: To make this work... remove this line.... and...

lib/MetaCPAN/Document/File.pm
```diff
@@ -363,6 +363,34 @@ sub _build_documentation {
...
+    lazy    => 1,
+    builder => '_build_suggest',
+);
+
+sub _build_suggest {
+    my $self = shift;
+    my $doc  = $self->documentation;
+    return +{} unless $doc;
```
ranguard: Change this to `return unless $doc`

PR #602, opened 2016-12-14
script/suggest: support day resolution slices for 'all'

mickeyn:
small change that was required while running the 'all' (backfilling for all files) - month resolution is not high enough (times out)


mickeyn:
@oalders: can you please check/merge this (fixed the tidy that made the tests fail) - this is a minor change to the suggest script.

PR #603, opened 2016-12-15
autocomplete_using_suggester: improve sorting by length

mickeyn:
sorting by length in the query to correct current results set.

### DIFF
--- a/lib/MetaCPAN/Document/File/Set.pm
+++ b/lib/MetaCPAN/Document/File/Set.pm
@@ -499,28 +499,48 @@ sub autocomplete_using_suggester {
         }
         );
 
-    my @docs
-        = map { $_->{text} } @{ $suggestions->{documentation}[0]{options} };
+    my %docs;
+
+    for my $suggest ( @{ $suggestions->{documentation}[0]{options} } ) {
+        next if exists $docs{ $suggest->{text} };
+        $docs{ $suggest->{text} } = $suggest->{score};
+    }
 
     my $data = $self->es->search(
         {
             index => $self->index->name,
             type  => 'file',
             body  => {
-                query  => { match_all => {} },
+                query => {
+                    filtered => {
+                        query => {
+                            function_score => {
+                                script_score => {
+                                    script => {
+                                        lang => 'groovy',
+                                        file =>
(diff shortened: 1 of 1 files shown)


lib/MetaCPAN/Document/File/Set.pm
```diff
@@ -499,28 +499,47 @@ sub autocomplete_using_suggester {
         }
         );
 
-    my @docs
-        = map { $_->{text} } @{ $suggestions->{documentation}[0]{options} };
+    my %docs;
+    for my $d ( @{ $suggestions->{documentation}[0]{options} } ) {
```
oalders: What is `$d`?  It's not immediately clear to me.  Maybe a more descriptive name?

mickeyn:
this change is for the not yet used suggester based autocomplete (for testing and comparison)

lib/MetaCPAN/Document/File/Set.pm
```diff
@@ -499,28 +499,47 @@ sub autocomplete_using_suggester {
...
+        next if exists $docs{ $d->{text} };
+        $docs{ $d->{text} } = $d->{score};
+    }
+
+    my @docs_uniq = sort { $docs{$b} <=> $docs{$a} } keys %docs;
```
oalders: Out of curiosity, why do these need to be sorted if it's being passed to Es in a query?

ranguard: ES does clever analysis, but can't order alphabetically on a suggest field!

oalders: Right, but it's being sorted on the way in rather than on the way out?

mickeyn: that's a redundant leftover from before adding the script ... will remove
I was looking for the reason it was getting low-quality results, and thought the `size => 10` which is clearly why we got random 10 matches from 'all' doc matches might be affected by the order... not smart :)

lib/MetaCPAN/Document/File/Set.pm
mickeyn: will rename

Issue #272, opened 2013-06-07
"Raw" access to images delivered as MIME="text/plain" instead of "image/png"


kentfredric:
If you look in the file listing for http://api.metacpan.org/source/KENTNL/MetaPOD-0.1.3/examples/ , you'll see it clearly knows the MIME type of the images in that dir:

``` text
Name    Size    Type    Last Modified
Parent Directory            
self_structure.dot  725 application/msword  Sun, 02 Jun 2013 08:28:08 GMT
self_structure.pl   859 text/x-script.perl  Sun, 02 Jun 2013 08:28:08 GMT
self_structure.png  27951   image/png   Sun, 02 Jun 2013 08:28:08 GMT
```

However, if you view the `png` file

http://api.metacpan.org/source/KENTNL/MetaPOD-0.1.3/examples/self_structure.png

It gets dispatched as `text/plain`

Which is not very useful.

haarg:
We should set X-Content-Type-Options: nosniff to prevent the files from being misused, which would prevent Chrome from displaying the image even when we are giving a different content type.

PR #605, opened 2016-12-20
use better content types for source requests

haarg:
If the API source endpoint is accessed directly, rather than through
st.aticpan.org, provide a more reasonable content type.  Images are
given a the appropriate type for the file, text files are served as
text/plain, and everything else is octet-stream.


haarg:
This fixes metacpan/metacpan-web#1748 and partly addresses #272.

PR #608, opened 2016-12-27
better link targets for headX/item

haarg:
Pod::Simple::XHTML does heavy filtering of the link targets it
generates.  This results in many targets being nearly useless for
external use.

Change the targets generated to use less filtering, but also generate
additional targets, one with the first word and one with the old target
like Pod::Simple::XHTML creates.

This is especially useful for pages like perlvar and perlfunc.  Many of
the targets in perlvar consist entirely of characters that are filtered
by Pod::Simple::XHTML.  The targets would therefore look like 'pod5'.
With this change, the unfiltered targets will be available, so linking
to variables will work properly.  The first word target is useful for
perlfunc or any other page that includes parameters function listings.
This will allow links like L<perlfunc/open> to work as people expect.

The Pod::Simple::XHTML style target is maintained for compatibility with
existing links on the web.

Unfortunately Pod::Simple::XHTML makes assumptions about the format of
the links, and has bugs with entity handling.  This works around these
and fixes some issues by encoding/decoding as appropriate using either
HTML entities or URL encoding.

perldoc_url_postfix handling is also fixed, even though we aren't using
it.

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -67,6 +67,7 @@ requires 'FindBin';
 requires 'Git::Helpers';
 requires 'Graph::Centrality::Pagerank';
 requires 'Gravatar::URL';
+requires 'HTML::Entities';
 requires 'HTML::TokeParser::Simple';
 requires 'HTTP::Request::Common';
 requires 'Hash::Merge::Simple';
--- a/lib/MetaCPAN/Pod/XHTML.pm
+++ b/lib/MetaCPAN/Pod/XHTML.pm
@@ -74,6 +74,7 @@ sub idify {
     $t = decode_entities($t);
     $t =~ s/^\s+//;
     $t =~ s/\s+$//;
+    $t =~ s/[\s-]+/-/g;
 
     return $self->url_encode($t)
         if $for_link;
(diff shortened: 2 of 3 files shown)


haarg:
One part of this that I'm unsure of is that the "canonical" targets generated maintain spaces in them, rather than filtering them to dashes.  And example is:
[https://metacpan.org/pod/distribution/perl/pod/perlfunc.pod#Perl-Functions-by-Category](https://metacpan.org/pod/distribution/perl/pod/perlfunc.pod#Perl-Functions-by-Category)
after this change, the target used in the index will be
[https://metacpan.org/pod/distribution/perl/pod/perlfunc.pod#Perl Functions by Category](https://metacpan.org/pod/distribution/perl/pod/perlfunc.pod#Perl%20Functions%20by%20Category).

haarg:
I've updated this to replace spaces with dashes in the link targets.

oalders:
`t/pod/renderer.t` is failing

haarg:
I've fixed the test.

oalders:
Thanks!

PR #610, opened 2017-01-11
Fix encoding of pod links

haarg:


haarg:
Fixes metacpan/metacpan-web#1842

PR #611, opened 2017-01-11
use classes rather than ids for pod error output

haarg:


oalders:
`t/server/controller/pod.t` as well as tidy are failing.

PR #615, opened 2017-02-04
Search testing

ranguard:
This is the initial work (that jberger did) wrapped up with carton+README - next stage is to convert it to use the API and then start trying to different weights, but wanted to get this merged in sooner rather than later


oalders:
👍

Issue #620, opened 2017-03-08
Add a flag to author to indicate that an author has passed away


oalders:
There was a suggestion that we should memorialize CPAN author pages.  In order to do this, the API needs to know about the deceased.  

There are two possible data sources:

`authors/00whois.xml`  In this file, the information is in the fullname key.  eg "<fullname>Nick Ing-Simmons (PAUSE Custodial Account)</fullname>"

There is also `Acme::CPANAuthors::InMemoriam` but it looks to lag behind `00whois.xml`.

@andk are there any cases of custodial accounts for authors who are still living or is this explicitly a way to flag an author as deceased?

mickeyn:
Maybe we shouldn't trust automation around this. but create a flag (in the authors type) that we only update manually upon request (and verification)?

oalders:
Are you saying not to trust automation because there might be some kind of typo in the custodial designation that would cause us to skip that author?  I believe that if it's flagged on the PAUSE side it has already gone through some kind of verification.

On the other hand, I guess anybody could update their full name to include that text.  That would be a weird edge case.

mickeyn:
i had 2 things in mind:
1. it doesn't (hopefully) happen often enough to deserve coding/maintenance.
2. it's a sensitive topic and we may not want any mistakes be made by code bugs.

the PAUSE verification is indeed all the verification we need, but we can monitor it for changes and just report them for manual updates rather than automatic.

that said, I don't have a strong objection here, just sharing the first things that came to mind.

andk:
>>>>> On Wed, 08 Mar 2017 10:51:49 -0800, Olaf Alders <notifications@github.com> said:

  > @andk are there any cases of custodial accounts for authors who are
  > still living or is this explicitly a way to flag an author as
  > deceased?

The practice of adding " (PAUSE Custodial Account)" has evolved circa 10
years ago and seems to have been applied consistently since ~2010. I
think it is reasonable to trust this. There are archives of
modules@perl.org to study who changed what when in case of doubt, but I
personally have no doubt. Thanks to those who cared so well for the
data.

-- 
andreas

shadowcat-mst:
@mickeyn it might not happen that often, but it's Not F-ing Fun and anything that makes the process easier would be gratefully received - after contacting pause and scouring for nickserv accounts to lock on magnet and freenode I'm generally Very Done with dealing it and just want to go get a drink.

oalders:
[...]

PR #622, opened 2017-03-27
Added 'custodial' flag to 'author' mapping.

mickeyn:
This flag will be set through Script::Author (bin/metacpan author).

To deploy we need to update the mapping of the current author type, running (on lw-mc-03):
`curl -X PUT 0:9200/cpan/_mapping/author -d'{"properties":{"is_pause_custodial_account":{"type":"boolean"}}}'`
**BEFORE** merging (can poke me to do it when we agree to merge)

This addresses #620

### DIFF
--- a/lib/MetaCPAN/Document/Author.pm
+++ b/lib/MetaCPAN/Document/Author.pm
@@ -92,6 +92,12 @@ has updated => (
     isa => 'DateTime',
 );
 
+has is_pause_custodial_account => (
+    is      => 'ro',
+    isa     => Bool,
+    default => 0,
+);
+
 sub _build_gravatar_url {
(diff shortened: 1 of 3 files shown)


oalders:
I wonder if we should call this `is_custodial_account`?  Or at least `custodial_account`?  The author object itself is not what's custodial.  Maybe `is_pause_custodial_account`?

mickeyn:
@oalders sure... many options :) I'll go with the last one (explicit ftw).

mickeyn:
@oalders do you want to merge this change?

oalders:
👍

mickeyn:
added the mapping. so should be updated on the next `author` script run.

PR #626, opened 2017-04-25
script author: use proper bulk + bug fix

mickeyn:
use bulk_helper instead of the model's bulk in the script (with longer timeout, being an hourly cron).

fix time comparison over the author's .json file's mtime - newer is '>' and not '>='.

### DIFF
--- a/lib/MetaCPAN/Document/Author.pm
+++ b/lib/MetaCPAN/Document/Author.pm
@@ -89,7 +89,7 @@ has extra => (
 
 has updated => (
     is  => 'ro',
-    isa => 'DateTime',
+    isa => Str,
 );
 
 has is_pause_custodial_account => (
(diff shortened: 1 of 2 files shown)


lib/MetaCPAN/Document/Author.pm
```diff
@@ -89,7 +89,7 @@ has extra => (
 
 has updated => (
     is  => 'ro',
-    isa => 'DateTime',
+    isa => 'Str',
```
oalders: We can remove the quotes around `Str`.

lib/MetaCPAN/Script/Author.pm
```diff
@@ -175,7 +187,7 @@ sub author_config {
         = { map { $_ => $author->{$_} }
             qw(name asciiname profile blog perlmongers donation email website city region country location extra)
         };
-    $author->{updated} = $mtime;
+    $author->{updated} = $mtime->strftime('%Y-%m-%dT%H:%M:%S');
```
oalders: Is this the same as `$mtime->iso8601`?

PR #628, opened 2017-05-08
Addde type 'packages' to support 02packages.details info

mickeyn:
Added mapping, document, script & tests to support adding
the contents of 02packages.details into its own new type
in the index (in a very similar way to 06perms).

Added modules:

* Document::Packages
document description of a package entry

* Script::Mapping::CPAN::Packages
type mapping for the index

* Script::Packages
a script to read the file & fill the index

* Server::Controller::Packages
an API controller to serve the data

Added test files:

* t/packages.t
* t/server/controller/packages.t


oalders:
Thanks @mickeyn!

PR #633, opened 2017-05-11
rename Packages --> Package

mickeyn:


mickeyn:
The second commit also added the distribution version extracted from the file value.

oalders:
I guess we'll need a `metacpan-puppet` PR to go along with this?

mickeyn:
right, https://github.com/metacpan/metacpan-puppet/pull/125

PR #636, opened 2017-05-12
Added an endpoint for looking up files for a release

mickeyn:
This endpoint /release/files/RELEASE?files=FILE1,FILE2,...
will lookup whether the release has files corresponding to
the given file names and will return the names & paths for
the found files.


oalders:
Looks good. Should we add a basic test for this?

mickeyn:
tests will be nice, but we'll add them once we add a test file for the release controller.
(in which case we'll go over and test all the release endpoints)

PR #637, opened 2017-05-12
Added new script and mapping for Debian package name mapping

mickeyn:
We want to provide the distribution name mappings for Perl
packages to 3rd party sources like Debian.

This commit adds the 'external_package' mapping, and starts
with adding the entry for 'debian' + the script to populate
the data.

### DIFF
--- a/lib/MetaCPAN/Script/Debian.pm
+++ b/lib/MetaCPAN/Script/Debian.pm
@@ -1,0 +1,181 @@
+package MetaCPAN::Script::Debian;
+
+use Moose;
+use namespace::autoclean;
+
+use CPAN::DistnameInfo;
+use DBI                   ();
+use Email::Sender::Simple ();
+use Email::Simple         ();
+use List::MoreUtils qw( uniq );
+
+use MetaCPAN::Types qw( HashRef Str );
+
+with 'MetaCPAN::Role::Script', 'MooseX::Getopt';
+
+has email_to => (
+    is       => 'ro',
+    isa      => Str,
+    required => 1,
+);
+
+has _host_regex => (
+    is      => 'ro',
+    isa     => Str,
+    lazy    => 1,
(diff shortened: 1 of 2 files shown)


lib/MetaCPAN/Script/Debian.pm
```diff
@@ -0,0 +1,182 @@
+package MetaCPAN::Script::Debian;
+
+use Moose;
+use namespace::autoclean;
+
+use DBI;
+use List::MoreUtils qw( uniq );
+use CPAN::DistnameInfo;
+use Email::Simple         ();
+use Email::Sender::Simple ();
```
oalders: Can we alpha sort these and not implicitly import?  ie `use DBI ()`;

lib/MetaCPAN/Script/Debian.pm
```diff
@@ -0,0 +1,182 @@
...
+    is       => 'ro',
+    isa      => Str,
+    required => 1,
+);
+
+has _host_rx => (
+    is      => 'ro',
+    isa     => Str,
+    lazy    => 1,
+    builder => '_build_host_rx',
```
oalders: How about `_build_host_regex`?

lib/MetaCPAN/Script/Debian.pm
```diff
@@ -0,0 +1,182 @@
...
+        # map Debian source package to CPAN distro
+        while ( my ( $source, $url ) = $sth->fetchrow ) {
+            next if $skip{$source};
+            $self->dist_for( $source, $url );
+            if ( my $dist = $self->dist_for( $source, $url ) ) {
+                $dist{$dist} = $source;
+            }
+            else {
+                push @failures => [ $source, $url ],
+                    ;
```
oalders: That semicolon should be on the previous line.  I think the extra comma is messing things up.

PR #639, opened 2017-05-12
Clean up search 

ranguard:
- refactor to make it cleaner
- lots of variable name updates for clarity
- move the logic that adds descriptions and favs into single place

IF this is ok then it should be easier to refactor from here

### DIFF
--- a/lib/MetaCPAN/Model/Search.pm
+++ b/lib/MetaCPAN/Model/Search.pm
@@ -11,11 +11,9 @@ use MetaCPAN::Types qw( Object Str );
 use MetaCPAN::Util qw( single_valued_arrayref_to_scalar );
 
 has es => (
-    is      => 'ro',
-    isa     => Object,
-    handles => {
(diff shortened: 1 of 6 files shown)


lib/MetaCPAN/Model/Search.pm
```diff
@@ -122,32 +138,45 @@ sub _search_collapsed {
...
+                count => { terms => { size => 999, field => 'distribution' } }
+                };
+
+        }
+
+        # Build the query and fetch results
+        my $es_query = $self->build_query( $search_term, $es_query_opts );
+        $es_results = $self->run_query( file => $es_query );
+
+        if ( $run == 1 ) {
```
jberger: was this some tidy thing?

ranguard: basically so it jumped up and slapped you in the face, rather than being a `..... if $run == 1;`

ranguard:
Close in favour of #646

PR #644, opened 2017-05-14
Added cleanup mode for script/permission

mickeyn:
This will allow cleaning up removed records once in a while
(it requires a longer run time so not suited for every cron run,
and records don't get deleted often anyways).

Usage: add --cleanup flag to the permission script execution.

### DIFF
--- a/lib/MetaCPAN/Script/Permission.pm
+++ b/lib/MetaCPAN/Script/Permission.pm
@@ -4,7 +4,8 @@ use Moose;
 
 use Log::Contextual qw( :log );
 use MetaCPAN::Document::Permission ();
-use PAUSE::Permissions             ();
+use MetaCPAN::Types qw( Bool );
+use PAUSE::Permissions ();
 
 with 'MooseX::Getopt', 'MetaCPAN::Role::Script';
 
@@ -14,6 +15,12 @@ Loads 06perms info into db. Does not require the presence of a local
 CPAN/minicpan.
 
 =cut
+
+has clean_up => (
+    is      => 'ro',
+    isa     => Bool,
+    default => 0,
+);
 
 sub run {
     my $self = shift;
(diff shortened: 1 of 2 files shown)


lib/MetaCPAN/Script/Permission.pm
```diff
@@ -6,6 +6,8 @@ use Log::Contextual qw( :log );
 use MetaCPAN::Document::Permission ();
 use PAUSE::Permissions             ();
 
+use MetaCPAN::Types qw( Bool );
```
oalders: Can we move this sorted into the other `use` statements?

lib/MetaCPAN/Script/Permission.pm
```diff
@@ -15,6 +17,12 @@ CPAN/minicpan.
 
 =cut
 
+has cleanup => (
```
oalders: Could we call this `clean_up`?

lib/MetaCPAN/Script/Permission.pm
```diff
@@ -28,40 +36,73 @@ sub index_permissions {
...
+        my $name = $perms->name;
+
         my @co_maints = $perms->co_maintainers;
         my $doc       = {
-            @co_maints
-            ? ( co_maintainers => \@co_maints )
-            : (),
-            module_name => $perms->name,
-            owner       => $perms->owner,
+            co_maintainers => \@co_maints,
```
oalders: Could we add a comment about why we're adding an empty array in some cases?

lib/MetaCPAN/Script/Permission.pm
```diff
@@ -28,40 +36,73 @@ sub index_permissions {
...
 
+sub run_cleanup {
+    my ( $self, $bulk, $seen ) = @_;
+
+    log_debug {"checking permission data to remove"};
+
+    my $scroll = $self->es->scroll_helper(
+        index  => $self->index->name,
+        type   => 'permission',
+        scroll => '30m',
```
oalders: Just wondering why the scroller needs to live for `30m`?

lib/MetaCPAN/Script/Permission.pm
```diff
@@ -28,40 +36,73 @@ sub index_permissions {
...
+        body   => { query => { match_all => {} } },
+    );
+
+    my @remove;
+    my $count = $scroll->total;
+    while ( my $p = $scroll->next ) {
+        my $id = $p->{_id};
+        unless ( exists $seen->{$id} ) {
+            push @remove => $id;
```
oalders: Could we use the standard comma here?

lib/MetaCPAN/Script/Permission.pm
mickeyn: takes 20m on my VM .... we'll run once a day - not an issue

PR #646, opened 2017-05-14
[WIP] improved search (esX::model issue to resolve)

ranguard:
This PR remains only as a reference point.

### DIFF
--- a/lib/MetaCPAN/Model/Search.pm
+++ b/lib/MetaCPAN/Model/Search.pm
@@ -11,11 +11,9 @@ use MetaCPAN::Types qw( Object Str );
 use MetaCPAN::Util qw( single_valued_arrayref_to_scalar );
 
 has es => (
-    is      => 'ro',
-    isa     => Object,
-    handles => {
-        _run_query => 'search',
-    },
+    is       => 'ro',
+    isa      => Object,
+    handles  => { _run_query => 'search', },
     required => 1,
 );
 
@@ -113,6 +111,8 @@ sub _search_expanded {
         = $self->_extract_and_inflate_results( $es_results, \@distributions );
 
     my $return = {
+
+        # results is array ref to be consistent with the collapsed version
         results   => [ map { [$_] } @$results ],
         total     => $es_results->{hits}->{total},
         collapsed => \0,
@@ -124,11 +124,11 @@ sub _search_collapsed {
     my ( $self, $search_term, $from, $page_size ) = @_;
 
     my $total;
+    my @distributions;
+    my $es_results;
+
     my $run  = 1;
     my $hits = 0;
-    my @distributions;
-    my $process_or_repeat;
-    my $es_results;
     do {
         # We need to scan enough modules to build up a sufficient number of
         # distributions to fill the results to the number requested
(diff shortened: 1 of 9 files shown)


lib/MetaCPAN/Model/Search.pm
```diff
@@ -1,23 +1,23 @@
...
-use Moose;
-
-use v5.10;
+use MetaCPAN::Moose;
 
+use Const::Fast qw( const );
 use Log::Contextual qw( :log :dlog );
-use MooseX::StrictConstructor;
 
+use Cpanel::JSON::XS;
```
oalders: We should explicitly import here.  `use Cpanel::JSON::XS qw( decode_json );`

lib/MetaCPAN/Model/Search.pm
```diff
@@ -1,23 +1,23 @@
...
 use Hash::Merge qw( merge );
 use List::Util qw( sum uniq );
 use MetaCPAN::Types qw( Object Str );
 use MetaCPAN::Util qw( single_valued_arrayref_to_scalar );
 
+# Read in here to share
```
oalders: Could we expand the comments here so that it makes sense without having to scan to the `__DATA__` section?

lib/MetaCPAN/Model/Search.pm
```diff
@@ -38,81 +38,108 @@ sub _not_rogue {
...
+        } @{ $es_results->{hits}->{hits} };
 
     # Everything after this will fail (slowly and silently) without results.
     return {} unless @distributions;
 
-    my @ids          = map { $_->{fields}->{id} } @{ $data->{hits}->{hits} };
-    my $descriptions = $self->search_descriptions(@ids);
-    my $favorites    = $self->search_favorites(@distributions);
-    my $results      = $self->_extract_results( $data, $favorites );
+    # Lookup favs and extract results from es (adding in favs)
```
oalders: s/Lookup/Look up/

lib/MetaCPAN/Model/Search.pm
```diff
@@ -368,6 +412,81 @@ sub build_query {
     return $search;
 }
 
+# This was an attempt at a totally new query but nees lots of
+# work to make the xt/search_web.t pass
+# May need a different querty for non-collapsed (anything
+# that uses lucen syndax, e.g. module:foo)
+sub build_query_attemt_at_new_query {
```
oalders: s/attemt/attempt/

lib/MetaCPAN/Model/Search.pm
```diff
@@ -368,6 +412,81 @@ sub build_query {
...
+    my ( $self, $search_term, $params ) = @_;
+    $params //= {};
+
+    #    ( my $clean = $search_term ) =~ s/::/ /g;
+    my $structure = decode_json $json;
```
oalders: It would be nice to use parens for functions that are not built-ins.

lib/MetaCPAN/Model/Search.pm
```diff
@@ -368,6 +412,81 @@ sub build_query {
...
+                    abstract.analyzed
+                    release
+                    path
+                    status
+                    indexed
+                    authorized
+                    distribution
+                    date
+                    id
+                    pod_lines
```
oalders: Any reason not to sort these?

oalders:
Should this be closed in favour of #741 ?

haarg:
This is long out of date, so would probably need to be recreated from scratch. I've renamed it so it's more obvious that this is attempting to improve the search query in case we want to revisit it. But for now this can be closed.

PR #650, opened 2017-05-14
Use the existing ua object

mickeyn:
We have a 'ua' object defined as attribute in the script role -
use it instead of creating one everywhere.

### DIFF
--- a/lib/MetaCPAN/Script/Release.pm
+++ b/lib/MetaCPAN/Script/Release.pm
@@ -10,7 +10,6 @@ BEGIN {
 use CPAN::DistnameInfo ();
 use File::Find::Rule;
 use File::stat ();
-use LWP::UserAgent;
 use Log::Contextual qw( :log :dlog );
 use MetaCPAN::Util;
 use MetaCPAN::Model::Release;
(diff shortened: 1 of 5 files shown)


lib/MetaCPAN/Script/Release.pm
```diff
@@ -112,15 +111,9 @@ sub run {
                 MetaCPAN::Util::author_dir( $d->cpanid ),
                 $d->filename,
             );
-            my $ua = LWP::UserAgent->new(
-                parse_head => 0,
```
oalders: This option isn't getting set in the `ua` builder.

lib/MetaCPAN/Script/Release.pm
```diff
@@ -112,15 +111,9 @@ sub run {
...
-                env_proxy  => 1,
-                agent      => 'metacpan',
-                timeout    => 30,
```
oalders: These options are also not set in the `ua` builder sub, with the exception of `env_proxy`.  I think it would at least be helpful to set the `agent`.  All of these have corresponding setter methods, so we could preserve them if we need to.

mickeyn:
@oalders fixed to address your comments

PR #654, opened 2017-05-17
Add a new role to collect cygwin-packaged CPAN distributions

book:


lib/MetaCPAN/Script/Role/External/Cygwin.pm
```diff
@@ -0,0 +1,72 @@
+package MetaCPAN::Script::Role::External::Cygwin;
+
+use v5.014;    # /r
```
mickeyn: this is not needed, we are running on perl 5.22

lib/MetaCPAN/Script/Role/External/Cygwin.pm
```diff
@@ -0,0 +1,72 @@
...
+);
+
+has mirror_timeout => (
+    is      => 'ro',
+    isa     => Int,
+    default => 10,
+);
+
+sub _build_mirrors {
+    my ($self) = @_;
```
mickeyn: for consistency - please use the same style ( shift | @_) as in the next method (or change that one)

mickeyn:
tidy check fail - please run tidyall on your VM (under the repo) and force push the updates

PR #658, opened 2017-05-18
Introducing 'contributor' type (release contributors info)

mickeyn:
New type and supporting code & test to fill/fetch the
release contributors info to/from the index.

* MetaCPAN::Script::Mapping::CPAN::Contributor
Added for type mapping for Elasticseach

* Script::Contributor
Added for data filling of release contributors.
Script can run in different modes:
--all          # run for all releases in the 'release' type
--distribution # run for all releases matching a given distribution
--release      # run for a single release (format: PAUSEID/DISTRIBUTION-VERSION)

* Script::Release
Updated to store contributor info for processed releases

* Script::Role::Contributor
Added to support common functionality shared by the scripts
for handling contributors info fetching/updating.

* Document::Contributor
Model representation of the new type.

* Server::Controller::Contributor
Controller endpoints for the new type.

### DIFF
--- a/lib/MetaCPAN/Script/Contributor.pm
+++ b/lib/MetaCPAN/Script/Contributor.pm
@@ -1,0 +1,114 @@
+package MetaCPAN::Script::Contributor;
+
+use strict;
+use warnings;
+
+use Moose;
+use Ref::Util qw( is_arrayref );
+
+use MetaCPAN::Types qw( Bool HashRef Str );
+
+with 'MetaCPAN::Role::Script', 'MooseX::Getopt',
+    'MetaCPAN::Script::Role::Contributor';
+
+has all => (
+    is            => 'ro',
+    isa           => Bool,
+    default       => 0,
+    documentation => 'update contributors for *all* releases',
+);
+
+has distribution => (
+    is  => 'ro',
+    isa => Str,
+    documentation =>
+        'update contributors for all releases matching distribution name',
+);
+
+has release => (
+    is  => 'ro',
+    isa => Str,
+    documentation =>
+        'update contributors for a single release (format: author/release_name)',
+);
+
+has author_release => (
+    is      => 'ro',
+    isa     => HashRef,
+    lazy    => 1,
+    builder => '_build_author_release',
+);
+
+sub _build_author_release {
+    my $self = shift;
+    return unless $self->release;
+    my ( $author, $release ) = split m{/}, $self->release;
+    $author && $release
+        or die
+        "Error: invalid 'release' argument (format: PAUSEID/DISTRIBUTION-VERSION)";
+    return +{
+        author  => $author,
+        release => $release,
+    };
+}
+
+sub run {
+    my $self = shift;
+
+    my $query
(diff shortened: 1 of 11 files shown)


mickeyn:
with a small change, I can make this new type live in its own isolated index... would you prefer this? (I think it's better as long as we don't need to share alias, which so far I don't think we do)

lib/MetaCPAN/Script/Contributor.pm
```diff
@@ -0,0 +1,118 @@
+package MetaCPAN::Script::Contributor;
+
+use strict;
+use warnings;
+
+BEGIN {
+    $ENV{PERL_JSON_BACKEND} = 'JSON::XS';
```
oalders: Can you remind me why we're doing this?

lib/MetaCPAN/Script/Role/Contributor.pm
```diff
@@ -0,0 +1,68 @@
+package MetaCPAN::Script::Role::Contributor;
+
+use Moose::Role;
+
+use MetaCPAN::Util qw< digest >;
+use Ref::Util qw< is_arrayref >;
```
oalders: Nitpick, but could we just stick with round brackets?

t/server/controller/contributor.t
```diff
@@ -0,0 +1,62 @@
+use strict;
+use warnings;
+
+use Cpanel::JSON::XS qw( decode_json );
+use MetaCPAN::Server::Test;
+use MetaCPAN::TestServer;
+use Test::More;
+
+my $server = MetaCPAN::TestServer->new;
+$server->index_packages;
```
oalders: `index_packages` is already called in `t/00_setup.t:`  Do we need to do it again here?

oalders:
> with a small change, I can make this new type live in its own isolated index... would you prefer this? (I think it's better as long as we don't need to share alias, which so far I don't think we do)

Works for me.  At the very least if there are bugs, this would be a good place to have them.

lib/MetaCPAN/Script/Contributor.pm
mickeyn: oops, here it was copy-pasted from Script::Release... there it was added in 2f6d07b5e2... not sure how relevant it is.
i'll remove it from this script.

lib/MetaCPAN/Script/Role/Contributor.pm
mickeyn: sure

t/server/controller/contributor.t
mickeyn: copy-pasted from the package test... not sure why it's there either - removing both.

mickeyn: oh, origin is t/server/controller/permission.t:
```
my $server = MetaCPAN::TestServer->new;
$server->index_permissions;
```

mickeyn:
@oalders I fixed as requested and updated to support separate index (makes more sense to me)
I created the new index & mapping - so it should be safe to merge.

PR #660, opened 2017-05-22
Added /release/requires/MODULE endpoint

mickeyn:
This new endpoint is meant to replace current Elasticsearch
query crafted in WEB and passed to the the /release/_search
endpoint.


mickeyn:
see: https://github.com/metacpan/metacpan-web/blob/master/lib/MetaCPAN/Web/Model/API/Module.pm#L84-L119

PR #665, opened 2017-05-28
Added Activity controller + /activity endpoint

mickeyn:
This new endpoint will replace the query currently generated and
sent from WEB.

### DIFF
--- a/lib/MetaCPAN/Document/Release.pm
+++ b/lib/MetaCPAN/Document/Release.pm
@@ -4,6 +4,7 @@ use strict;
 use warnings;
 
 use Moose;
+use DateTime qw();
 use Ref::Util qw();
 use ElasticSearchX::Model::Document;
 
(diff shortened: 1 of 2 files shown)


lib/MetaCPAN/Document/Release.pm
```diff
@@ -591,5 +592,77 @@ sub requires {
     };
 }
 
+sub _activity_filters {
+    my ( $self, $params, $start ) = @_;
+    my ( $author, $distribution, $module, $f )
```
oalders: Could we clarify what `$f` is?  It's not clear to me from the name or the context of the code.  :)

lib/MetaCPAN/Document/Release.pm
```diff
@@ -591,5 +592,77 @@ sub requires {
...
+        = @{$params}{qw( author distribution module f )};
+
+    my @filters
+        = ( { range => { date => { from => $start->epoch . '000' } } } );
```
oalders: Could we add a comment as to what the `000` suffix does?

lib/MetaCPAN/Document/Release.pm
mickeyn: this is from the original code in WEB, dating back to 2012, i'll rename to `$new_dists`

lib/MetaCPAN/Document/Release.pm
mickeyn: this is from the original code in WEB, dating back to 2011 - first commit to the activity controller... of course no explanation there :)

oalders: Heh, ok.  :)

PR #666, opened 2017-05-28
Improved /author/by_user endpoint

mickeyn:
In WEB use, the query runs against multiple user ids.

This change first fixes the query to support multi-value check
(term -> terms).

Added another support for this endpoint which takes the users'
ids through a 'user' URL param: 'user=USER_ID1&user=USER_ID2...'

If you don't see a need for the URL param support I can drop it (I found it useful for manual testing using curl but WEB will probably not need it).

### DIFF
--- a/lib/MetaCPAN/Server/Controller/Author.pm
+++ b/lib/MetaCPAN/Server/Controller/Author.pm
@@ -61,9 +61,20 @@ sub get : Path('') : Args(1) {
         ['The requested field(s) could not be found'] );
 }
 
+# /author/by_user/USER_ID
 sub by_user : Path('by_user') : Args(1) {
     my ( $self, $c, $user ) = @_;
     my $data = $self->model($c)->raw->by_user($user);
+    $data or return;
+    $c->stash($data);
+}
+
+# /author/by_user?user=USER_ID1&user=USER_ID2...
+sub by_users : Path('by_user') : Args(0) {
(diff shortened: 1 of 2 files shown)


lib/MetaCPAN/Server/Controller/Author.pm
```diff
@@ -61,9 +62,26 @@ sub get : Path('') : Args(1) {
...
+    $c->stash($data);
+}
+
+# GET:  /author/by_user?users=USER_ID1,USER_ID2...
+# POST: /author/by_user  {"users":["USER_ID1","USER_ID2"...]}
+sub by_users : Path('by_user') : Args(0) {
+    my ( $self, $c ) = @_;
+
+    # multiple user ids in 'users' param(s) (can be CSV)
+    my $users = $c->req->params->{users} // $c->req->body_data->{users};
```
oalders: How about just using `my @users = $c->req->param('users')` here?  That should cover both `GET` and `POST` and we'd be guaranteed to get a list, so no need to check `is_arrayref` below.

https://metacpan.org/pod/Catalyst::Request#$req->param

Also, I don't think we need the split since we could send `user=foo&user=bar&user=baz` to get the same result.

mickeyn:
@oalders is it good to be merged?

oalders:
Thanks @mickeyn!

PR #667, opened 2017-05-28
favorite: set by_user default to match WEB usage

mickeyn:
WEB is running this query with a fixed 1000 value... no point in setting the default to a different one.


mickeyn:
actually, there are 4 calls 2 with size=1000 and 2 using the default... so i'll leave it as is for now

oalders:
Sorry I missed this!

mickeyn:
it's good you did... i didn't want it merged eventually :)

PR #668, opened 2017-05-29
favorite: by_user: added missing author details

mickeyn:


mickeyn:
this is a tiny change with no side-effects. will merge as soon as tests are finished.

PR #672, opened 2017-05-30
Fix type updates in distribution

mickeyn:
This PR is for 2 issues with scripts that both update 'distribution'

1. River is making destructive updates, deleting the external_package data.
2. External is not creating new records.


mickeyn:
@book you may want to review the changes to External.pm

book:
External.pm looks ok to me.

PR #679, opened 2017-06-03
Added /favorite/leaderboard API endpoint

mickeyn:
This new endpoint will be used to replace the query sent from
WEB.

### DIFF
--- a/lib/MetaCPAN/Server/Controller/Favorite.pm
+++ b/lib/MetaCPAN/Server/Controller/Favorite.pm
@@ -45,5 +45,12 @@ sub recent : Path('recent') : Args(0) {
     $c->stash($data);
 }
 
+sub leaderboard : Path('leaderboard') : Args(0) {
+    my ( $self, $c ) = @_;
+    my $data = $self->model($c)->raw->leaderboard();
+    $data or return;
+    $c->stash($data);
+}
+
 __PACKAGE__->meta->make_immutable;
 1;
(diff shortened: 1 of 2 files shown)


lib/MetaCPAN/Server/Controller/Favorite.pm
```diff
@@ -45,5 +45,12 @@ sub recent : Path('recent') : Args(0) {
     $c->stash($data);
 }
 
+sub leaderboard : Path('leaderboard') : Args(0) {
+    my ( $self, $c ) = @_;
+    my $data = $self->model($c)->raw->leaderboard();
+    $data or return;
+    $c->stash($data);
```
oalders: In general, I wonder if we should just shorten these to:

`return $c->stash( $self->model($c)->raw->leaderboard || {} );`

Not sure if you can stash an `undef`.

oalders:
@haarg this build appears to have failed bec. Carton is trying to get `provides` via the `v0` endpoint.

lib/MetaCPAN/Server/Controller/Favorite.pm
mickeyn: you can't stash a non-hashref value ... that's where this check comes from.
i don't mind the suggested change, but i assigned $data for debugging purposes.

mickeyn:
the tests passed now, but the v0 error is still thrown in the setup test

PR #694, opened 2017-06-21
Added 'total' & 'took' to /release/interesting_files endpoint

mickeyn:
These values will simplify adoption by WEB


mickeyn:
this is the same change as in #692 so i'll go ahead and merge it

PR #701, opened 2017-06-28
Added /favorite/agg_by_distirbutions API endpoint

mickeyn:
This new endpoint will replace query sending from WEB.

### DIFF
--- a/lib/MetaCPAN/Document/Favorite.pm
+++ b/lib/MetaCPAN/Document/Favorite.pm
@@ -122,6 +122,59 @@ sub users_by_distribution {
     return { users => \@plusser_users };
 }
 
+sub agg_by_distributions {
+    my ( $self, $distributions, $user ) = @_;
+    return unless $distributions;
+
+    my $body = {
+        size  => 0,
+        query => {
+            terms => { 'distribution' => $distributions }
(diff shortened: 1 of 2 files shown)


lib/MetaCPAN/Document/Favorite.pm
```diff
@@ -122,6 +122,59 @@ sub users_by_distribution {
     return { users => \@plusser_users };
 }
 
+sub agg_by_distirbutions {
```
oalders: s/distirbutions/distributions/

lib/MetaCPAN/Server/Controller/Favorite.pm
```diff
@@ -59,5 +59,26 @@ sub leaderboard : Path('leaderboard') : Args(0) {
     $c->stash($data);
 }
 
+sub agg_by_distirbutions : Path('agg_by_distributions') : Args(0) {
```
oalders: s/distirbutions/distributions/

lib/MetaCPAN/Document/Favorite.pm
mickeyn: nice catch

PR #705, opened 2017-06-29
/favorite/agg_by_distributions: change result structure

mickeyn:
For easier adoption by WEB, the returned results are better in
hashes rather than arrays.


mickeyn:
This is basically matching the output with the current method in WEB issuing the query, so i'm going to go ahead and merge it.

PR #708, opened 2017-07-04
/release/modules: no single-value collapsing on 'modules'

mickeyn:
Keep single-value collapsing on data from 'fields', not from
'_source'.

This will fix:
19:40:20 � haarg� something happened to our linking on release pages
19:40:55 � haarg� https://metacpan.org/release/Moo they are using file paths rather than /pod/Module

### DIFF
--- a/lib/MetaCPAN/Document/Release.pm
+++ b/lib/MetaCPAN/Document/Release.pm
@@ -1148,8 +1148,10 @@ sub modules {
     );
     return unless $ret->{hits}{total};
 
-    my @files = map { single_valued_arrayref_to_scalar($_) }
-        map +{ %{ $_->{fields} }, %{ $_->{_source} } },
+    my @files = map +{
+        %{ ( single_valued_arrayref_to_scalar( $_->{fields} ) )[0] },
+        %{ $_->{_source} }
+        },
         @{ $ret->{hits}{hits} };
 
     return {


lib/MetaCPAN/Document/Release.pm
```diff
@@ -1148,9 +1148,10 @@ sub modules {
     );
     return unless $ret->{hits}{total};
 
-    my @files = map { single_valued_arrayref_to_scalar($_) }
-        map +{ %{ $_->{fields} }, %{ $_->{_source} } },
-        @{ $ret->{hits}{hits} };
+    my @files = map {
+        single_valued_arrayref_to_scalar( $_->{fields} );
+        +{ %{ $_->{fields} }, %{ $_->{_source} } }
+    } @{ $ret->{hits}{hits} };
```
haarg: Possibly
```
map {
    +{
        %{ single_valued_arrayref_to_scalar( $_->{fields} ) },
        %{ $_->{_source} },
    }    
} @{ $ret->{hits}{hits} };
```
instead?  I don't like relying on the modification of a parameter.

mickeyn: that's what I tried first, but it produces an error:
Can't use string ("1") as a HASH ref ...

this code was a quick working compromise.

haarg:
```
    %{ (single_valued_arrayref_to_scalar( $_->{fields} ))[0] },
```
should work.

Alternatively, single_valued_arrayref_to_scalar should be changed from
```
    return $is_arrayref ? $array : @{$array};
```
to
```
    return $is_arrayref ? $array : $array->[0];
```

mickeyn: ok, the first option works - let's go with this one.
I rather deal with single_valued_arrayref_to_scalar itself later (half its code isn't even used so we can simplify it).

I'll push a fix.

PR #711, opened 2017-07-05
Added /search/web/first API endpoint

mickeyn:
This will simplify the reading of /search/web/simple output by
extracting the first element and eliminating the need for WEB
to read an ES output structure.

### DIFF
--- a/lib/MetaCPAN/Model/Search.pm
+++ b/lib/MetaCPAN/Model/Search.pm
@@ -44,6 +44,16 @@ sub search_simple {
     return $results;
 }
 
+sub search_for_first_result {
+    my ( $self, $query ) = @_;
+    my $es_query = $self->build_query($query);
+    my $results = $self->run_query( file => $es_query );
+    return unless $results->{hits}{total};
+    my $data = $results->{hits}{hits}[0];
+    single_valued_arrayref_to_scalar( $data->{fields} );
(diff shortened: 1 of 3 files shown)


lib/MetaCPAN/Model/Search.pm
```diff
@@ -44,6 +44,16 @@ sub search_simple {
     return $results;
 }
 
+sub search_first {
```
oalders: Why is this called `search_first`?  Can we give it a more descriptive name or add a comment?

mickeyn: oh, that name was copied of search_simple & search_web.... it's basically search_simple but only returns the first element (so web doesn't have to dig through {hits}{hits} structures.

I don't mind changing it, any suggestions?

oalders: In that case maybe even just `search_for_first_result` would convey what it's doing a little more clearly?

oalders: Might be helpful to add a note explaining when or why we want to do this.

mickeyn: i'll change the name as suggested.
just to be clear, it's the only usage of search_simple... so technically can be used only as a temp. name until we remove the need for search_simple

PR #714, opened 2017-07-07
Moved all Document::Set packages to their own files

mickeyn:
don't like having multiple packages in the same files + they are getting too long anyway.


lib/MetaCPAN/Document/Author/Set.pm
```diff
@@ -0,0 +1,112 @@
+package MetaCPAN::Document::Author::Set;
+
+use strict;
+use warnings;
+
+use Moose;
```
oalders: Are these classes able to use `MetaCPAN::Moose` or does the ESX model get in the way?

mickeyn: MetaCPAN::Moose is not used anywhere in API (not even in its cpanfile)... we can try switching to it, but that should be a separate commit (I'll start playing with it soon).
These packages were simply broken off as is from the Document files.

PR #719, opened 2017-07-10
Added 400/404s for Release endpoints

mickeyn:


mickeyn:
@haarg if that's what you had in mind, i can quickly apply to the rest of the controllers

haarg:
Yeah, this looks good to me.

mickeyn:
requires #720 to pass the mirrors test

PR #728, opened 2017-07-31
/reverse_dependencies: added took+total info

mickeyn:


mickeyn:
this change doesn't break any functioality, just adds info to an endpoint output... so merging.

PR #730, opened 2017-08-22
Added missing cdn_max_age setting

mickeyn:


lib/MetaCPAN/Server.pm
```diff
@@ -150,10 +150,13 @@ sub read_param {
...
     my ( $c, $data ) = @_;
-    $data
-        ? $c->stash($data)
-        : $c->detach( '/not_found',
-        ['The requested info could not be found'] );
+    if ($data) {
+        $c->stash($data);
+    }
+    else {
+        $c->cdn_stale_if_error('1M');    # taken from Root::end()
```
ranguard: Before reading below... please confirm if `$c->detach('/not_found')` actually calls the end() in root....

If it does - you don't need the $c->cdn_stale... line above.

If it does NOT:

You need the whole logic:
```
    if ( $c->cdn_max_age ) {

        # If we allow caching, we can serve stale content, if we error
        # on backend. Because we have caching on our UI (metacpan.org)
        # we don't really want to use stale_while_revalidate on
        # our API as otherwise the UI cacheing could be of old content
        $c->cdn_stale_if_error('1M');
    }
    else {
        # Default to telling fastly NOT to cache unless we have a
        # cdn cache time
        $c->cdn_never_cache(1);
    }
```
This way if something that uses `stash_or_detach` and hasn't set a cdn_max_age we tell Fastly to NOT cache the response in anyway what so ever.

mickeyn: I don't see that it does... actually, I don't see any calls to end()... is that being called implicitly by the framework?

I'll add the logic then.

PR #731, opened 2017-09-01
more api refactoring

mickeyn:
moved some more data handling code out of the controllers into the document modules.

once merged, at least one new endpoint (release.by_author_and_name) could be cleaned out (after WEB switches to 'get')

### DIFF
--- a/lib/MetaCPAN/Document/File/Set.pm
+++ b/lib/MetaCPAN/Document/File/Set.pm
@@ -3,6 +3,7 @@ package MetaCPAN::Document::File::Set;
 use Moose;
 
 use MetaCPAN::Util qw( single_valued_arrayref_to_scalar );
+use Ref::Util qw( is_hashref );
 
 extends 'ElasticSearchX::Model::Document::Set';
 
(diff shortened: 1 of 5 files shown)


lib/MetaCPAN/Document/File/Set.pm
```diff
@@ -703,5 +703,67 @@ sub interesting_files {
...
+    # find the most likely file
+    # TODO: should we do this when the release is indexed
+    # and store the result as { 'changes_file' => $name }
+
+    my @candidates = qw(
+        CHANGES Changes ChangeLog Changelog CHANGELOG NEWS
+    );
+
+    # use $c->model b/c we can't let any filters apply here
+    my $file = eval {
```
oalders: Why is this in an `eval`?

lib/MetaCPAN/Document/Release/Set.pm
```diff
@@ -9,6 +9,34 @@ use MetaCPAN::Util qw( single_valued_arrayref_to_scalar );
 
 extends 'ElasticSearchX::Model::Document::Set';
 
+sub author_status {
+    my ( $self, $id, $file ) = @_;
+    return unless $id and $file;
+
+    my $st = $file->{_source}
```
oalders: It's not clear to me what `$st` is.  Is this `status`?  If so, could we use that name?  :)

oalders: I see this is a cut/paste, but might be a good time to clean up that variable name.

lib/MetaCPAN/Document/File/Set.pm
mickeyn: cut/paste

lib/MetaCPAN/Document/Release/Set.pm
mickeyn: sure. I'll do some cleaning up

oalders:
@mickeyn thanks! BTW, the build is failing.

mickeyn:
hmm... that test doesn't fail on my VM :) ... will have to find a way to debug it.

mickeyn:
@oalders are you OK with merging this one?

oalders:
Yes, thanks @mickeyn!

PR #732, opened 2017-09-13
[WIP] (incomplete) test file for the release controller

mickeyn:


mickeyn:
@oalders this is still incomplete as far as the controller goes, but I think it's a good idea to merging it to have a couple more endpoints tested

PR #741, opened 2017-10-26
Leo/search refactor 2.0 - a cleanup

ranguard:
This is mostly just renaming variables that were confusing and adding some pod here and there. There isn't any change to actual queries.

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -30,7 +30,6 @@ requires 'CatalystX::RoleApplicator';
 requires 'CatalystX::Fastly::Role::Response', '0.06';
 requires 'CPAN::Repository::Perms';
 requires 'Config::ZOMG', '>=', '1.000000';
-requires 'Config::JFDI';
 requires 'Const::Fast';
 requires 'Cpanel::JSON::XS', '3.0115';
 requires 'Cwd';
--- a/cpanfile
+++ b/cpanfile
@@ -30,6 +30,8 @@ requires 'CatalystX::RoleApplicator';
 requires 'CatalystX::Fastly::Role::Response', '0.06';
 requires 'CPAN::Repository::Perms';
 requires 'Config::ZOMG', '>=', '1.000000';
+requires 'Config::JFDI';
+requires 'Const::Fast';
 requires 'Cpanel::JSON::XS', '3.0115';
 requires 'Cwd';
 requires 'Data::Printer', '0.38';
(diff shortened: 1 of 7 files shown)


cpanfile
```diff
@@ -30,6 +30,8 @@ requires 'CatalystX::RoleApplicator';
 requires 'CatalystX::Fastly::Role::Response', '0.06';
 requires 'CPAN::Repository::Perms';
 requires 'Config::ZOMG', '>=', '1.000000';
+requires 'Config::JFDI';
```
oalders: I *think* we should no longer be using `Config::JFDI` as of an earlier merge, so this can be removed.

t/model/search.t
```diff
@@ -0,0 +1,108 @@
...
+use warnings;
+
+use MetaCPAN::Model::Search ();
+use MetaCPAN::TestServer    ();
+use Test::More;
+use Test::Deep;
+
+plan skip_all =>
+    "Travis ES bad, see https://travis-ci.org/metacpan/metacpan-api/jobs/301092129"
+    if $ENV{TRAVIS};
```
oalders: Do we know why this is happening on Travis?

t/model/search.t
```diff
@@ -0,0 +1,108 @@
+use strict;
+use warnings;
+
+use MetaCPAN::Model::Search ();
+use MetaCPAN::TestServer    ();
+use Test::More;
+use Test::Deep;
```
oalders: `use Test::Deep qw(cmp_deeply);`

lib/MetaCPAN/Model/Search.pm
```diff
@@ -1,9 +1,8 @@
 package MetaCPAN::Model::Search;
 
-use Moose;
-
-use v5.10;
+use MetaCPAN::Moose;
```
oalders: 👍

lib/MetaCPAN/Model/Search.pm
```diff
@@ -38,91 +35,123 @@ sub _not_rogue {
...
+        $search_term,
         {
             size => $page_size,
             from => $from
         }
     );
 
     #return $es_query;
-    my $data = $self->run_query( file => $es_query );
+    my $es_results = $self->run_query( file => $es_query );
```
oalders: 👍

PR #742, opened 2017-10-26
WIPLeo/will this smoke ok

ranguard:
just testing looking at https://travis-ci.org/metacpan/metacpan-api/builds this seems to have been the last successful build https://travis-ci.org/metacpan/metacpan-api/builds/270730811


ranguard:
strange looks like data - but we don't use anything external do we?

ranguard:
not needed now

PR #746, opened 2017-11-08
[DO NOT MERGE YET] Reverse dependencies: read sort option from body

mickeyn:
The 'sort' parameter is passed from WEB as JSON in a POST request
as it is a complex structure (hash).

This change requires this merge in WEB first: https://github.com/metacpan/metacpan-web/pull/1981


mickeyn:
closing in favor of a better solution

PR #762, opened 2017-11-17
add log4perl logging to server web app

haarg:
The scripts use their own configuration for now, although preferrably we
will merge them together.

### DIFF
--- a/lib/MetaCPAN/Server/Controller/User.pm
+++ b/lib/MetaCPAN/Server/Controller/User.pm
@@ -5,6 +5,7 @@ use warnings;
 
 use DateTime;
 use Moose;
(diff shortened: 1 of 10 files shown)


lib/MetaCPAN/Server/Controller/User.pm
```diff
@@ -5,6 +5,7 @@ use warnings;
 
 use DateTime;
 use Moose;
+use Log::Log4perl::MDC;
```
ranguard: ? - why is this needed here?

haarg: Later in the file it is used to attach some extra request data to the logging.

PR #763, opened 2017-11-17
Created MetaCPAN::Query::*

mickeyn:
This new namespace will be used to hold ES query code that
is not coupled with the model and will eventually replace
the model.

As a first step, moving the contents of MetaCPAN::Document::Favorite::Set
to the new namespace.
This will be used to serve a shared access for the autocomplete method.

### DIFF
--- a/lib/MetaCPAN/Query/Role/Common.pm
+++ b/lib/MetaCPAN/Query/Role/Common.pm
@@ -1,0 +1,11 @@
+package MetaCPAN::Query::Role::Common;
+
+use Moose::Role;
+use MetaCPAN::Types qw( Str );
+
+has es => ( is => 'ro', );
+
+has index_name => ( is => 'ro', );
+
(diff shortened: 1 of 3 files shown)


lib/MetaCPAN/Query/Role/Common.pm
```diff
@@ -0,0 +1,11 @@
+package MetaCPAN::Query::Role::Common;
+
+use Moose::Role;
+use MetaCPAN::Types qw( Str );
+
+has es => ( is => 'ro', );
+
+has index_name => ( is => 'ro', );
```
oalders: Do these attributes still need an `isa` and maybe `required`?

mickeyn: will add that later... just needed that merged for Joel's coming suggester changes that depend on this change.

PR #765, opened 2017-11-17
add script to read data from cpantesters api

preaction:
This script reads the release test summary data from the CPAN Testers
API (http://api.cpantesters.org) instead of the SQLite database (which
has been prone to problems).

Once this is merged, we'll have to change over the update process to use this script, and then once this process proves stable, we can remove the old script.


oalders:
LGTM. Just wondering why there's so much logic in a shell script? :)

preaction:
> why there's so much logic in a shell script?

[I learned it from you, okay?! I learned it from watching you!!!!!!!!1](https://github.com/metacpan/metacpan-api/blob/6444e26dca81f3457c6f80a63354838c2d6bdcd8/bin/cpantesters_mini_db_for_testing)

But no, that was a pain. I'm 100% open to suggestions. It's mostly like that so that adding new modules is, like the other cpantesters fake db bin script, easy to do with a single line. I could change it such that it was just a Perl script that fetched the thing and only wrote the things we wanted instead...

PR #767, opened 2017-11-17
query namespace for shared model actions

mickeyn:
more non-model related querying code moved to the Query namespace


mickeyn:
@oalders done

PR #771, opened 2017-11-18
change badge (png -> svg) and add kritika badge

skaji:
before
screen shot 2017-11-18 at 9 53 08

after
screen shot 2017-11-18 at 17 41 20

Thank you all for meta::hack v2!


oalders:
Thanks @skaji!

PR #773, opened 2017-11-18
Added a Favorite script

mickeyn:
This script will be used to fill dist_fav_count in 'file' type
(to boost search abilities)

The script supports queueing for parallel distribution handling.

### DIFF
--- a/lib/MetaCPAN/Script/Favorite.pm
+++ b/lib/MetaCPAN/Script/Favorite.pm
@@ -1,0 +1,183 @@
+package MetaCPAN::Script::Favorite;
+
+use Moose;
+
+use Log::Contextual qw( :log );
+
+use MetaCPAN::Types qw( Bool Int Str );
+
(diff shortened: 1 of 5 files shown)


lib/MetaCPAN/Script/Favorite.pm
```diff
@@ -0,0 +1,183 @@
...
+    }
+    else {
+        my $favs = $self->es->scroll_helper(
+            index       => $self->index->name,
+            type        => 'favorite',
+            search_type => 'scan',
+            scroll      => '30s',
+            fields      => [qw< distribution >],
+            size        => 500,
+            ( body => $body ) x !!$body,
```
oalders: Is this a different way of saying `$body ? ( body => $body) : (),`

mickeyn: yes, that's the "enterprise" way :)

mickeyn:
@oalders I added the script run to 00setup for now

Issue #572, opened 2016-11-20
add 'deprecated' field to release type


mickeyn:
can be taken from the meta, let's add it to the release result as a query-able field 

It should also look at the POD `name` attribute and look for `DEPRECIATED`  or `DEPRECATED` (case sensitive)

oalders:
I think we should add this field to modules as well.

oalders:
See also https://github.com/metacpan/metacpan-web/issues/1435

mickeyn:
This is now done, both `file` and `release` types have a `deprecated` field.

PR #781, opened 2018-01-17
Added 'deprecated' field

mickeyn:
Added to both release and file types (which is also /module in API)
Field will be set by the 'release' script.

This PR is to address #572 

Note: after merge - need to run a one-off run to update field for existing releases

### DIFF
--- a/lib/MetaCPAN/Document/File.pm
+++ b/lib/MetaCPAN/Document/File.pm
@@ -32,6 +32,21 @@ sub BUILD {
 
 =head1 PROPERTIES
 
+=head2 deprecated
+
+Indicates file deprecation (the abstract contains "DEPRECATED" or "DEPRECIATED",
+or the x_deprecated flag is included in the corresponding "provides" entry in distribution metadata);
+it is also set if the entire release is marked deprecated (see L<MetaCPAN::Document::Release#deprecated>).
+
+=cut
+
+has deprecated => (
+    is      => 'ro',
+    isa     => Bool,
+    default => 0,
+    writer  => '_set_deprecated',
+);
+
 =head2 abstract
 
 Abstract of the documentation (if any). This is built by parsing the
--- a/lib/MetaCPAN/Document/Release.pm
+++ b/lib/MetaCPAN/Document/Release.pm
@@ -104,6 +104,14 @@ This is an ArrayRef of modules that are included in this release.
 
 =cut
 
+=head2 deprecated
+
+This is a boolean indicating whether the release is marked as deprecated
+(the main module's ABSTRACT contains "DEPRECATED" or "DEPRECIATED",
+or the x_deprecated flag is set in metadata)
+
+=cut
+
 has provides => (
     is     => 'ro',
     isa    => ArrayRef [Str],
@@ -241,6 +249,13 @@ has changes_file => (
     is     => 'ro',
     isa    => Str,
     writer => '_set_changes_file',
+);
+
+has deprecated => (
+    is      => 'ro',
+    isa     => Bool,
+    default => sub {0},
+    writer  => '_set_deprecated',
 );
 
 sub _build_download_url {
--- a/lib/MetaCPAN/Script/Release.pm
+++ b/lib/MetaCPAN/Script/Release.pm
@@ -237,6 +237,15 @@ sub import_archive {
             = [ @{ $associated_pod{$documentation} || [] }, $_ ];
     }
 
+# check for release deprecation in abstract of release or has x_deprecated in meta
+    my $deprecated = (
+               $meta->{x_deprecated}
+            or $document->has_abstract
+            and $document->abstract =~ /DEPRECI?ATED/
+    ) ? 1 : 0;
+
+    $document->_set_deprecated($deprecated);
+
     log_debug { 'Indexing ', scalar @$modules, ' modules' };
     my $perms = $self->perms;
     my @release_unauthorized;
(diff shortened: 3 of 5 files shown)


lib/MetaCPAN/Document/File.pm
```diff
@@ -32,6 +32,20 @@ sub BUILD {
...
+
+Indicates file deprecation (this will normally mean the abstract will contain
+DEPRECATED / DEPRECIATED or the release is marked with x_deprecated = 1
+
+=cut
+
+has deprecated => (
+    is      => 'ro',
+    isa     => Bool,
+    default => sub {0},
```
oalders: You should also be able to set a `Bool` default via `default => 0`,

lib/MetaCPAN/Document/Release.pm
```diff
@@ -104,6 +104,12 @@ This is an ArrayRef of modules that are included in this release.
 
 =cut
 
+=head2 deprecated
+
+This is a boolean indicated whether the release is marked as deprecated
```
oalders: s/indicated/indicating/

lib/MetaCPAN/Script/Release.pm
```diff
@@ -254,6 +263,12 @@ sub import_archive {
         $file->clear_module if ( $file->is_pod_file );
         $file->documentation;
         $file->suggest;
+
+        # check for DEPRECATED/DEPRECIATED in abstract of file
+        $file->_set_deprecated(1)
+            if $deprecated
+            or $file->abstract and $file->abstract =~ /DEPRECI?ATED/;
```
oalders: +1 for searching for the misspelling.

lib/MetaCPAN/Script/Release.pm
```diff
@@ -237,6 +237,15 @@ sub import_archive {
...
     }
 
+# check for release deprecation in abstract of release or has x_deprecated in meta
+    my $deprecated = (
+               $meta->{x_deprecated}
+            or $document->has_abstract
+            and $document->abstract =~ /DEPRECI?ATED/
+    ) ? 1 : 0;
+
+    $document->_set_deprecated($deprecated);
```
oalders: First off, thanks so much for doing this.  I have wanted to see something like this for a long time.  My only real issue with the logic is that it seems to blur the lines between a release which is deprecated and a release which contains only *some* deprecated modules.  See https://metacpan.org/pod/release/ETHER/Dist-Zilla-Plugin-Deprecated-0.007/lib/Dist/Zilla/Plugin/Deprecated.pm#DESCRIPTION

Basically, checking for presence of the key will not always be quite enough.  If a release has 10 modules and one of them has been deprecated using `x_deprecated` key, then it looks like the entire release will be flagged.

lib/MetaCPAN/Script/Release.pm
oalders: Based on the above comments, I think we need a bit more logic around flagging files as having been deprecated.  Are there any corner cases here we should note, like files containing inner packages where some could theoretically have been deprecated but others not?  Not sure how likely this is, but if it's possible, then it's worth noting.

lib/MetaCPAN/Script/Release.pm
mickeyn: OK, that clarifies the issue I ran into with this :)
I'll try to fix the logic.

mickeyn: This one is indicating a release deprecation, i.e. the x_deprecated is in the top level of meta.
A single deprecated file does not set this flag, in that case we will have the x_deprecated key set under `meta->provides->file->x_deprecated` so that logic is just missing for the file later on - will add that.

mickeyn: added the mentioned check on file level

lib/MetaCPAN/Script/Release.pm
mickeyn: ok, so now the file's check for x_deprecated is in place... this is still a check for any of the file's modules being deprecated as we are marking the file and not a single module (I guess that's a fair compromise in this case).

So other files will not be affected (check using: `bin/metacpan release --latest https://cpan.metacpan.org/authors/id/E/ET/ETHER/App-Nopaste-1.011.tar.gz` - see only `Codepeek.pm` gets deprecated) 

The open question I have is - do we want to mark all the file entries as deprecated when the release is marked as deprecated?

oalders: > The open question I have is - do we want to mark all the file entries as deprecated when the release is marked as deprecated?

I think so.

oalders:
@karenetheridge please feel free to review/comment on this.

lib/MetaCPAN/Document/File.pm
```diff
@@ -32,6 +32,20 @@ sub BUILD {
 
 =head1 PROPERTIES
 
+=head2 deprecated
+
+Indicates file deprecation (this will normally mean the abstract will contain
```
karenetheridge: Will the file get the 'deprecated' flag if the entire release is deprecated, as well as the individual file is deprecated? If so, the documentation isn't quite right: it would be something more like:

    Indicates file deprecation (the abstract contains "DEPRECATED" or "DEPRECIATED", or the x_deprecated flag is included in the corresponding "provides" entry in distribution metadata); it is also set if the entire release is marked deprecated (see L<MetaCPAN::Document::Release#deprecated>).

lib/MetaCPAN/Document/Release.pm
```diff
@@ -104,6 +104,12 @@ This is an ArrayRef of modules that are included in this release.
 
 =cut
 
+=head2 deprecated
+
+This is a boolean indicating whether the release is marked as deprecated
+
```
karenetheridge: I would add:

   ... (the main module's ABSTRACT contains "DEPRECATED" or "DEPRECIATED", or the x_deprecated flag is set in metadata)

lib/MetaCPAN/Document/File.pm
mickeyn: added as suggested

lib/MetaCPAN/Document/Release.pm
mickeyn: added as suggested

mickeyn:
thanks @karenetheridge

mickeyn:
updated the mapping in production - now it's safe to merge

PR #782, opened 2018-02-02
less noise from script

mickeyn:
Can we please make that script less noisy?


preaction:
This and #783 should fix the problem.

PR #784, opened 2018-02-03
Script::CPANTestersAPI - increase bulk_helper timeout

mickeyn:
increase the bulk helper timeout. 

@preaction do you think 30m is enough?


preaction:
The request to CPAN Testers will take like 2-5 minutes, I suspect. I have no idea how long updating the data in Elastic will take, but 30 minutes sounds good.

PR #790, opened 2018-04-06
Map values of 'deprecated' to 0,1 to ensure correct sorting

mickeyn:
This is because I've seen cases where ES returns the quoted
strings "true" & "false" where in others the JSON true & false.

Normalizing to 0 & 1 ensures the correct sorting we do based
on this field in the results from the autocomplete suggester.

### DIFF
--- a/lib/MetaCPAN/Document/File/Set.pm
+++ b/lib/MetaCPAN/Document/File/Set.pm
@@ -580,6 +580,12 @@ sub autocomplete_suggester {
         ( $_->{fields}{documentation}[0] => \%record );
     } @{ $data->{hits}{hits} };
 
+    # normalize 'deprecated' field values to boolean (1/0) values (because ES)
+    for my $v ( values %valid ) {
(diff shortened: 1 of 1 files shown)


lib/MetaCPAN/Document/File/Set.pm
```diff
@@ -580,6 +580,12 @@ sub autocomplete_suggester {
         ( $_->{fields}{documentation}[0] => \%record );
     } @{ $data->{hits}{hits} };
 
+    # normalize 'deprecated' field values to boolean (1/0) values (because ES)
+    for my $v ( values %valid ) {
+        $v->{deprecated} = 1 if $v->{deprecated} =~ /^true$/;
+        $v->{deprecated} = 0 if $v->{deprecated} =~ /^false$/;
```
haarg: Couldn't these just be `eq` checks?

mickeyn: yeah, actually it can - will fix

PR #793, opened 2018-04-19
Change queries to use search type 'dfs_query_then_fetch'

mickeyn:
This will hopefully fix the issue of sometimes badly ordered
results in search.


mickeyn:
this is only for Search... not all queries

PR #802, opened 2018-04-21
Standalone API/endpoint for finding a release's CONTRIBUTING/HACKING file

zakame:
This implements a function and endpoint similar to the current Changes API, but for searching a release's CONTRIBUTING file, if available.  This also searches for HACKING files (which serve a similar purpose, especially on older releases,) and `perlhack.pod` for Perl releases.

Needed to support https://github.com/metacpan/metacpan-web/issues/1779.

### DIFF
--- a/lib/MetaCPAN/Document/File/Set.pm
+++ b/lib/MetaCPAN/Document/File/Set.pm
@@ -667,5 +667,42 @@ sub find_changes_files {
     return $file->{_source};
 }
 
+sub find_contributing_files {
+    my ( $self, $author, $release ) = @_;
+
+    my @candidates = qw(
+        CONTRIBUTING CONTRIBUTING.md Contributing.pod Contributing
+        HACKING HACKING.md HACKING.pod Hacking.pod Hacking
+    );
+
+    my $file = $self->raw->filter(
+        {
+            and => [
+                { term => { release => $release } },
+                { term => { author  => $author } },
+                {
+                    or => [
+                        {
+                            and => [
+                                { term => { distribution => 'perl' } },
+                                { term => { name         => 'perlhack.pod' } }
+                            ]
+                        },
+                        {
+                            or => [
+                                map { { term => { name => $_ } } }
+                                    @candidates
+                            ]
+                        }
+                    ]
+                }
+            ]
+        }
+    )->size(1)->first;
+
+    return unless is_hashref($file);
+    return $file->{_source};
+}
+
 __PACKAGE__->meta->make_immutable;
 1;
--- a/lib/MetaCPAN/Server/Controller/ContributingDoc.pm
+++ b/lib/MetaCPAN/Server/Controller/ContributingDoc.pm
@@ -25,4 +25,6 @@ sub get : Chained('index') : PathPart('') : Args(2) {
     $c->stash($file);
 }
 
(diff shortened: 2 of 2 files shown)


oalders:
Why not just add `HACKING HACKING.md HACKING.pod Hacking.pod Hacking` to the `interesting_files` endpoint?  That's essentially what they are and it would avoid the overhead of maintaining more code, some of which is duplicated elsewhere.

zakame:
Thanks! Yes, I considered just adding to `MetaCPAN::Query::File->interesting_files`, but found it not as useful:

- it didn't take into account `Contributing.pod` or `Hacking.pod` that's in `lib/`, unless I remove `{ term => { level => 0 } }` from the filter; also, the `perlhack` special case
- it gave me a list of files, most of which wasn't really needed for generating a single link; I could perhaps do `List::Util::first { /$contributing_or_hacking_re/ } @interesting_files` on the metacpan-web controller side, but that seems wasteful, whereas ES should be able to give only one result

I do dislike the apparent duplication to `find_changes_files` though, so I'll look into refactoring this or both functions (they basically share logic, only their input lists and ES filters are different.)

oalders:
> it didn't take into account Contributing.pod or Hacking.pod that's in lib/, unless I remove { term => { level => 0 } } from the filter

So why not remove that from the filter?

> but that seems wasteful, whereas ES should be able to give only one result

That may be more than you need, but there's a human cost to every additional line of code.  This project has a lot of complexity, a lot of code and very few active contributors.  If we can cover more than one case with the same API query, then that's a win.

I commented previously that I'd like to continue to use the existing API, but this code does the opposite of that.  I want to work with you to get your code merged, but part of that process is either accepting guidance from the core contributors or changing our minds., even if you don't agree.  I don't see a compelling argument here for what you're proposing.  We're talking about saving a few bytes.  That's a micro-optimization that isn't justified in this case.

zakame:
> So why not remove that from the filter?

I did, and that seems to work, but I do not know what impact would that have on any other cases (e.g. on some module with a big-enough file list, like Paws or App-NetDisco; there's probably a test case needed here.

> I commented previously that I'd like to continue to use the existing API, but this code does the opposite of that. I want to work with you to get your code merged, but part of that process is either accepting guidance from the core contributors or changing our minds., even if you don't agree. I don't see a compelling argument here for what you're proposing. We're talking about saving a few bytes. That's a micro-optimization that isn't justified in this case.

Perhaps I may have misunderstood you at https://github.com/metacpan/metacpan-web/issues/1779 when I asked advice about adding the new API call or not; did you actually mean `I would just add the extra call (if needed)` to be a call to `interesting_files` instead of a new API?  Because if so, I would have preferred that you were more specific on that context. (Furthermore, your reaction in https://github.com/metacpan/metacpan-web/issues/1779#issuecomment-383305192 indicated to me that my progress to implement this new call is ok...)

On a lighter note,

> accepting guidance from the core contributors or changing our minds

Isn't this what we're already doing now? :grin:

zakame:
@oalders Alright, I think I got another implementation that just modifies `interesting_files` per your intent, in https://github.com/metacpan/metacpan-api/pull/803.

I'll close this one, and change https://github.com/metacpan/metacpan-web/pull/2042 to use the newer PR.  Thanks!

PR #803, opened 2018-04-22
Add more CONTRIBUTING/HACKING docs as interesting files

zakame:
This enables https://github.com/metacpan/metacpan-web/issues/1779 by letting `interesting_files` pick up Contributing files as well as Hacking files on older dists, and relaxes the filter to let the function search for these files (and more) throughout the release dist.

### DIFF
--- a/lib/MetaCPAN/Query/File.pm
+++ b/lib/MetaCPAN/Query/File.pm
@@ -63,7 +63,6 @@ sub interesting_files {
                                 {
                                     bool => {
                                         must => [
-                                            { term => { level => 0 } },
                                             {
                                                 terms => {
                                                     name => [
@@ -74,6 +73,9 @@ sub interesting_files {
                                                             CHANGES
                                                             CONTRIBUTING
                                                             CONTRIBUTING.md
+                                                            CONTRIBUTING.pod
+                                                            Contributing.pm
+                                                            Contributing.pod
                                                             COPYING
                                                             COPYRIGHT
                                                             CREDITS
@@ -82,6 +84,12 @@ sub interesting_files {
                                                             Changes
                                                             Copying
                                                             FAQ
+                                                            HACKING
+                                                            HACKING.md
+                                                            HACKING.pod
+                                                            Hacking.pm
+                                                            Hacking.pod
+                                                            Hacking
                                                             INSTALL
                                                             INSTALL.md
                                                             LICENCE


oalders:
@zakame I apologize for the delay.  I was just wrapping up my time at PTS when you sent this and then I had to travel for $work.  [Insert more lame excuses here].

Could you squash these into one commit?  My guess is that the test failure has been handled by now and the build should pass on the next run.

lib/MetaCPAN/Query/File.pm
```diff
@@ -63,7 +63,6 @@ sub interesting_files {
                                 {
                                     bool => {
                                         must => [
-                                            { term => { level => 0 } },
```
oalders: Regarding this change here, can you remind me why we need to do this?  Will these new files not generally be at the top level?

zakame: Yep in https://github.com/metacpan/metacpan-api/pull/803/commits/78bfcb6d2fc2280683f1df75452fb2dc446ff967

I wanted to match places like https://metacpan.org/pod/distribution/Moose/lib/Moose/Manual/Contributing.pod as well, hence this change.

oalders: Ah, that makes sense to me.  @mickeyn does that work for you?  I don't know of any performance implications, but maybe you do?

zakame: Yeah, let me know if there's a better/more maintainable way :muscle:

zakame:
Hi @oalders! Just squashed this per https://github.com/metacpan/metacpan-web/pull/2042#issuecomment-387522706, thanks!

oalders:
Thanks @zakame!

lib/MetaCPAN/Query/File.pm
Grinnz: Given the filenames matched, this seems like it should be ok, but I'm still wary. Will the pod file in question be displayed as 'Moose/Manual/Contributing.pod' for example if a distribution has several such files throughout the hierarchy?

zakame: Yes, Moose::Manual::Contributing still shows up under Documentation

image

Grinnz: I think that removing this from the query has caused the problems I was worried about, where it picks up various internal copies that are not as interesting as the one at root level. Example: https://metacpan.org/release/Dist-Zilla#other

oalders: Maybe we just want to filter the results to prefer the file which is closest to the root of the directory?

zakame: @Grinnz @oalders thanks, I'll look into this and try to get an PR out by the weekend :muscle:

PR #804, opened 2018-04-22
scripts updates

mickeyn:
Bug fix for Script::Mapping and a few code updates, mainly to help with debugging of Script::Latest


oalders:
@mickeyn this needs to be tidied.

PR #805, opened 2018-04-22
Added river data per distribution(s) endpoints

mickeyn:


mickeyn:
eek, I see an editor cache file in there... i'll clean it out

oalders:
@mickeyn I think there's an issue here.  This PR included a file called `lib/MetaCPAN/Document/Distribution/Set.pm~`.  Maybe we should delete it?

oalders:
Ah, I see you've dealt with it.  :)

PR #817, opened 2018-05-28
Add Changes.md to interesting files

robrwo:
As per @oalder's suggestion in
https://github.com/metacpan/metacpan-web/issues/2043


haarg:
The logic between the important files and the picking the change log specifically don't match.  This will show Changes.md files in the root in the list of interesting files, but for the change log for a release it will use a Changes.md file in a subdirectory, or and of the other forms (Changelog.md etc).  I think for .md files, we should stick to only the files that exist in the root of the repo.

haarg:
Implemented in 9e9aac8525706d6b25baf971ec3b0144d2048a94.

PR #819, opened 2018-06-11
Minor changes to make the API docs more accessible

jmaslak:
Some developers that may use this API may not be familiar with terms like API and DSL, or even why you would care about an API.


oalders:
Thanks @jmaslak!

PR #826, opened 2018-06-16
clear author's user when disconnecting from PAUSE

haarg:
Untested.

### DIFF
--- a/lib/MetaCPAN/Model/User/Account.pm
+++ b/lib/MetaCPAN/Model/User/Account.pm
@@ -146,5 +146,24 @@ sub get_identities {
     return grep { $_->name eq $identity } @{ $self->identity };
 }
 
+sub remove_identity {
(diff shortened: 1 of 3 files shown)


lib/MetaCPAN/Model/User/Account.pm
```diff
@@ -146,5 +146,24 @@ sub get_identities {
     return grep { $_->name eq $identity } @{ $self->identity };
 }
 
+sub remove_identity {
+    my ($self, $identity ) = @_;
+    my $ids = $self->identities;
+    my ($id) = grep { $_->{name} eq $identity } @$ids;
+    @$ids = grep { $_->{name} ne $identity } @$ids;
```
mickeyn: I think you could do one iteration of map instead of 2 greps

oalders:
Looks like the tidyall tests are failing.

PR #829, opened 2018-06-20
Remove leading './' from path given in META to match dir listing

mickeyn:
When parsing a release we compare the paths of files in 'provides'
with the directory listing.
When these entries in META contain './' they don't match the listing.
This causes the 'modules' section in the 'release' entry to miss
files and in turn (if 'modules' ends up empty) the release cannot
be marked as 'latest'.

@oalders @haarg I know this will fix a case like https://github.com/metacpan/metacpan-web/issues/2074, but wonder if it can actually cause issues.
Your thoughts?


oalders:
> @oalders @haarg I know this will fix a case like metacpan/metacpan-web#2074, but wonder if it can actually cause issues.
Your thoughts?

I can't think of an issue.  I'm inclined to deploy it and see if anything breaks.  ;)

mickeyn:
sounds good. will merge and then we can try to reindex any release that isn't marked as 'latest' but should.

PR #831, opened 2018-06-25
use new Search::Elasticsearch and ElasticSearchX::Model

haarg:
This seems to be working OK, but could use some additional testing, especially in the scripts.


oalders:
Does this PR need more tests or should we merge?

haarg:
I'd appreciate it if people could do some manual testing of the scripts.

mickeyn:
@haarg I'll try to do some live testing tonight

mickeyn:
tested a few scripts manually, seems to be working fine, merging

haarg:
Had to revert this.  The script code isn't using values from the config file to connect to ES, and isn't loading the config files correctly.

PR #836, opened 2018-08-02
Update the DSN to the UDD

book:
The address and credentials have changed. See https://udd-mirror.debian.net/ for details.


mickeyn:
only fails the tidy test, will merge and then push a fix (to speed it up)

PR #838, opened 2018-08-06
Referenced URL unresponsive

jonasbn:
The URL does not respond on https, but http.


jonasbn:
Well would love to correct it in the other end, but I do not have the proper privileges ;)

PR #876, opened 2018-11-19
Create Model, Controller and spec for Cover

ssoriche:
Using `MetaCPAN::Model::User` and `MetaCPAN::Model::Download` as examples, convert `MetaCPAN::Query::Cover` into a model, add the plugin, controller, and tests, and create the OpenAPI spec to drive it.

### DIFF
--- a/lib/MetaCPAN/API/Model/Cover.pm
+++ b/lib/MetaCPAN/API/Model/Cover.pm
@@ -1,0 +1,31 @@
+package MetaCPAN::API::Model::Cover;
+
+use MetaCPAN::Moose;
+
+with 'MetaCPAN::API::Model::Role::ES';
+
+sub find_release_coverage {
+    my ( $self, $release ) = @_;
+
+    my $query = +{ term => { release => $release } };
+
+    my $res = $self->_run_query(
+        index => 'cover',
+        type  => 'cover',
+        body  => {
+            query => $query,
+            size  => 999,
+        }
+    );
+    $res->{hits}{total} or return {};
+
+    return +{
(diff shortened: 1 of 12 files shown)


lib/MetaCPAN/API/Model/Cover.pm
```diff
@@ -0,0 +1,33 @@
...
+    my $res = $self->_run_query(
+        index => 'cover',
+        type  => 'cover',
+        body  => {
+            query => $query,
+            size  => 999,
+        }
+    );
+    use DDP;
+    p $res;
```
oalders: Leftover debugging?

oalders:
This is failing on `t/api/controller/cover.t` in Travis.

lib/MetaCPAN/API/Model/Cover.pm
ssoriche: Oops, trying to do too many things at one time. Cleaning it out in an additional commit that I’ll squash after approval before it gets merged.

mickeyn:
@ssoriche, the test probably fails because there is no data indexed into the `cover` type during `t/0*`.

t/api/controller/cover.t
```diff
@@ -4,6 +4,7 @@ use Test::More;
 use Test::Mojo;
 use Mojo::JSON qw(true false);
 
+$ENV{ES} = 'localhost:9300';
```
mickeyn: what's on :9300 ? do we have a new ES instance?

ssoriche:
Yes please. I have a lot of cleanup to do in this branch to remove some "noise".

ssoriche:
Should be good now.

Issue #327, opened 2014-08-10
Provide checksums for files in download_url (of distribution)


rehsack:
Please add support for fetching checksums via MetaCPAN API to allow verification of downloaded sources. Interesting checksums are for SHA1 and RDN160 for Pkgsrc and MD5 and SHA256 for Yocto. Generally it would be cool when the checksum method could be specified and metacpan.org could use an own mirror for on-the-fly generation.

mikroskeem:
Any updates on this?

oalders:
Looks like this is just waiting for a volunteer?  I think it would be a great feature to have.  I read this the other day and it had me thinking more about checksums: https://blog.golang.org/modules2019

oalders:
Closed via #883.  Thanks @mickeyn!

PR #883, opened 2018-12-27
Added checksums (md5/sha256) to release archives + download_url endpoint

mickeyn:
Relates to #327 

This will add `checksum_md5` & `checksum_sha1` mappings *definitions* (will require manually adding the mappings in production - I can take care of that once approved) to the `release` type (I don't think the `file` type level is the right place to hold it though it has the `download_url` copy as it may confuse people to think these are the checksums of the files).

Is `Digest::file` the right module to use? any other suggestions?

### DIFF
--- a/lib/MetaCPAN/Document/File/Set.pm
+++ b/lib/MetaCPAN/Document/File/Set.pm
@@ -8,6 +8,7 @@ use List::Util qw( max );
 
 use MetaCPAN::Query::File;
 use MetaCPAN::Query::Favorite;
+use MetaCPAN::Query::Release;
 
 extends 'ElasticSearchX::Model::Document::Set';
 
@@ -38,6 +39,22 @@ has query_favorite => (
 sub _build_query_favorite {
     my $self = shift;
     return MetaCPAN::Query::Favorite->new(
+        es         => $self->es,
+        index_name => $self->index->name,
+    );
+}
+
+has query_release => (
+    is      => 'ro',
+    isa     => 'MetaCPAN::Query::Release',
+    lazy    => 1,
+    builder => '_build_query_release',
+    handles => [qw< get_checksums >],
+);
+
+sub _build_query_release {
+    my $self = shift;
+    return MetaCPAN::Query::Release->new(
         es         => $self->es,
         index_name => $self->index->name,
     );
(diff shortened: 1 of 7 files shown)


ranguard:
We should take the checksum from those provided in the CPAN dir if possible (I guess fall back for backpan if needed)?

mickeyn:
@ranguard I don't see a full coverage of all authors (or releases) under http://www.cpan.org/modules/by-module/CPAN/
Do you know if there is one? otherwise, I don't think we should bother with a solution that will only work for a (small?) subset of the cases.

mikroskeem:
Why md5 or sha1? Why not sha256 or others?

oalders:
@andk is there an authoritative list of checksums for all CPAN uploads?

ranguard:
> @ranguard I don't see a full coverage of all authors (or releases) under http://www.cpan.org/modules/by-module/CPAN/
> Do you know if there is one? otherwise, I don't think we should bother with a solution that will only work for a (small?) subset of the cases.

@mickeyn All authors should have a CHECKSUMS in their author DIR.. e.g..

https://www.cpan.org/modules/by-authors/id/M/MI/MICKEY/CHECKSUMS
https://www.cpan.org/modules/by-authors/id/L/LL/LLAP/CHECKSUMS

These are for the latest version of releases they have done I believe

https://pause.perl.org/pause/query?ACTION=pause_05news suggests it uses https://metacpan.org/pod/CPAN::Checksums to generate the CHECKSUMS file

The reason I suggest using the one from PAUSE where possible is that it then means if WE (metacpan) had an issue syncing the file we don't serve an MD5/sha based on a corrupted file.

Though I understand it will be more work so *shrug* - something better than nothing. And the likelyhood of corrupted between us and upstream is slim.

oalders:
I'd prefer to get it from PAUSE as well, but this does only cover current releases, so if we want BackPAN coverage we'll need some sort of combined solution.

mickeyn:
I'm OK with the CPAN solution as long as we can cover all releases with it (few exceptions are acceptable as long as we don't need to maintain a separate code for it but fix them manually or something).

@oalders for the old releases we could potentially use `CPAN::Checksums` to generate the missing data (like @ranguard suggested) - this is mostly a one-off so can be scripted for back-filling.

I think, in this case, i'd prefer it as a separate index rather than added columns to the `release` index.

andk:
@oalders: the checksums files are not per-author but per-directory. As far as I know, they never reached any sort of systematic backup. Sorry!

PR #888, opened 2019-03-07
Log::Any via Log::Log4perl

haarg:
Log::Any is used by Search::Elasticsearch, so we should connect it to
our standard logger.


oalders:
I think this PR will need the `cpanfile` and `cpanfile.snapshot` changes.

PR #892, opened 2019-04-11
Take several (delayed) passes at setting the "latest" flag

oalders:
The previous behaviour was to run this job immediately after the
indexing job finished, but this was proabably too early in almost all
cases.  Here we wait for a while and then try setting the flag.


oalders:
I'm not sure if the tests touch this code path.

PR #893, opened 2019-04-12
Move email sending to a model class

oalders:
This relies on the credentials PR I just created getting merged first.


oalders:
I need to add some more tests before I request a review.

oalders:
This should be ready for review now.  I checked and I think the test coverage is actually ok.

PR #894, opened 2019-04-18
Set th "size" of resultset to be the same as input size.

gugod:
Otherwise Elasticsearch returns 10 hits by default.  Since this
'by_author_and_names' routine is the backend of '/feed/recent' and
responds up to 100 hits, this tweak can reduce the amount of roundtrip
bteewn frontend and backend.

I wish to utilize this together with https://github.com/metacpan/metacpan-web/pull/2157


mickeyn:
thanks!

PR #914, opened 2019-04-28
clean up some prereqs and unused scripts

haarg:


oalders:
The one thing I did notice about `bin/prove` was that setting `export METACPAN_SERVER_CONFIG_LOCAL_SUFFIX=testing` was necessary to get the tests passing on Docker.

haarg:
Travis sets the environment variables itself.

oalders:
I was more thinking of the Docker env.  Maybe we can get the test suite to DTRT in both places.

PR #935, opened 2019-12-06
Added checksum backfilling script

mickeyn:


lib/MetaCPAN/Script/Checksum.pm
```diff
@@ -0,0 +1,120 @@
+package MetaCPAN::Script::Checksum;
+
+use Moose;
```
oalders: We could also use `MetaCPAN::Moose` here, I think.

mickeyn: none of the scripts use it (yet), should be part of a separate change

PR #944, opened 2019-12-18
Skip indexing distributions in a Raku subdir

Grinnz:
To allow for future use of a Raku subdir, as in https://github.com/andk/pause/issues/346


oalders:
Thanks @Grinnz!

PR #953, opened 2020-04-07
escape reserved chars in search term

### LINKED ISSUE #952: Search queries which include a slash trigger an exception
https://fastapi.metacpan.org/search/web?q=perl%2Ftk&size=50&from=0

mickeyn:
fixes #952 - ES reserved chars in non-simple string search require escaping.

(needs tidy)


oalders:
See https://stackoverflow.com/questions/40384936/elasticsearch-query-string-fails-to-parse-query-with-some-characters

PR #956, opened 2020-04-18
Fix author/dist/module searches

oalders:
This is a quick fix for now.  It might be helpful to have a larger discussion about how to treat colons, but this would restore the status quo while keeping the fix for slashes in searches in place.


oalders:
This should fix most of the failing front end tests.

PR #959, opened 2020-06-13
convert uses of JSON to Cpanel::JSON::XS

haarg:


oalders:
We can fix the tidy failures later.

PR #964, opened 2020-06-17
Search fix

mohawk2:
The previous fix (https://github.com/metacpan/metacpan-api/commit/225749b6e54a29ee7a34b835d01f0918e7448b7f) had two issues:
- it matched spaces, so it quoted them - I was unable to make it not do this, so I removed all the whitespace and the `x` flag
- it only quoted the first character it found, not all

This is hopefully a fix, and hopefully will also fix the -web tests.

### DIFF
--- a/lib/MetaCPAN/Model/Search.pm
+++ b/lib/MetaCPAN/Model/Search.pm
@@ -79,18 +79,15 @@ sub search_web {
     $page_size //= 20;
     $from      //= 0;
 
-    $search_term
-        =~ s{([ + - = > < ! & | ( ) { } \[ \] ^ " ~ * ? \ / ])}{\\$1}x;
+    $search_term =~ s{([+=><!&|\(\)\{\}[\]\^"~*?\\/])}{\\$1}g;
 
     # munge the search_term
     # these would be nicer if we had variable-length lookbehinds...
     # Allow q = 'author:LLAP' or 'module:Data::Page' or 'dist:'
-    # We are mapping to correct ES fields here - wonder if ANYONE
-    # uses these?!?!?!
+    # We are mapping to correct ES fields here - relied on by metacpan-web
+    # tests.
     #
     # The exceptions below are used specifically by the front end search.
-    # We've temporarily removed the ":" from the regex above so that the the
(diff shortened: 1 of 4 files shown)


t/model/search.t
```diff
@@ -59,6 +59,17 @@ ok( $search->_not_rogue, '_not_rogue' );
...
 }
 
+{
+    my $results = $search->search_web('author:Mo BadPod');
+    isnt( @{ $results->{results} },
+        0, '>0 results on author search with extra' );
+}
+
+{
+    eval { $search->search_web('usr/bin/env') };
```
oalders, line 69: We also could have used `Test::Fatal` here. Not sure if that's much of a problem, though.

lib/MetaCPAN/Model/Search.pm
```diff
@@ -79,18 +79,15 @@ sub search_web {
     $page_size //= 20;
     $from      //= 0;
 
-    $search_term
-        =~ s{([ + - = > < ! & | ( ) { } \[ \] ^ " ~ * ? \ / ])}{\\$1}x;
+    $search_term =~ s{([+=><!&|\(\)\{\}[\]\^"~*?\\/])}{\\$1}g;
```
oalders, line 82: Good detective work!  🔍

t/model/search.t
mohawk2: I was aiming for no extra deps ;-)

lib/MetaCPAN/Model/Search.pm
mohawk2: I thought making it not match spaces would be easy. In fact, after half an hour I gave up and just zapped the whitespace!

t/model/search.t
oalders: :) It must be a secondary dep, since we do already use it: https://github.com/metacpan/metacpan-api/blob/master/t/api/controller/admin.t#L5

oalders: Although, look at that test, I'm not sure we're actually making use of it there.

mohawk2: It wasn't - zapped in another commit at your request :-)

PR #965, opened 2020-06-17
Partial switch to Type::Tiny

mohawk2:
The CI ran about 10% quicker (about 9:30 vs 10:30) compared to last one, though that may be a fluke.


oalders:
Thanks!!!

PR #971, opened 2020-06-28
Show prereqs files on release page if present

Grinnz:
https://metacpan.org/pod/Dist::Zilla::Plugin::PrereqsFile


oalders:
Thanks @Grinnz!

Issue #927, opened 2019-11-23
Feature request: add an endpoint that returns perl tarball urls in plain text format


skaji:
metacpan has an endpoint that returns perl tarball urls in json format:
https://fastapi.metacpan.org/v1/release/versions/perl

It's very helpful, thanks.

On the other hand, the json format is not very easy-to-use by shell scripts.
In fact, I'm thinking about creating a perl-installer in shell script.

So it's nice metacpan has an endpoint that returns perl tarball urls in plain text format.

My expected response is something like:
```
5.31.6     \t https://cpan.metacpan.org/authors/id/B/BI/BINGOS/perl-5.31.6.tar.gz \n
5.30.1     \t https://cpan.metacpan.org/authors/id/S/SH/SHAY/perl-5.30.1.tar.gz \n
5.30.1-RC1 \t https://cpan.metacpan.org/authors/id/S/SH/SHAY/perl-5.30.1-RC1.tar.gz \n
5.31.5     \t https://cpan.metacpan.org/authors/id/S/SH/SHAY/perl-5.31.5.tar.gz \n
5.31.4     \t https://cpan.metacpan.org/authors/id/C/CO/CORION/perl-5.31.4.tar.gz \n
5.31.3     \t https://cpan.metacpan.org/authors/id/T/TO/TOMHUKINS/perl-5.31.3.tar.gz \n
5.31.2     \t https://cpan.metacpan.org/authors/id/S/SH/SHAY/perl-5.31.2.tar.gz \n
5.31.1     \t https://cpan.metacpan.org/authors/id/E/ET/ETHER/perl-5.31.1.tar.gz \n
5.31.0     \t https://cpan.metacpan.org/authors/id/X/XS/XSAWYERX/perl-5.31.0.tar.gz \n
5.30.0     \t https://cpan.metacpan.org/authors/id/X/XS/XSAWYERX/perl-5.30.0.tar.gz \n
...
```

Additionally, it would be nice we can change the response with some query parameter:

* Returns stable perl tarball urls only (say, with query `?version=stable`)
    ```
    5.30.1 \t https://cpan.metacpan.org/authors/id/S/SH/SHAY/perl-5.30.1.tar.gz \n
    5.30.0 \t https://cpan.metacpan.org/authors/id/X/XS/XSAWYERX/perl-5.30.0.tar.gz \n
    5.28.2 \t https://cpan.metacpan.org/authors/id/S/SH/SHAY/perl-5.28.2.tar.gz \n
[...]

PR #988, opened 2020-10-20
GH#927 Add plain mode + version filter to release/versions endpoint

mickeyn:
This PR is addressing #927 

---

This commit adds support for two new parameters to the release/versions endpoint.

Examples:

https://fastapi.metacpan.org/v1/release/versions/perl?plain=1
https://fastapi.metacpan.org/v1/release/versions/perl?versions=5.26.0,5.30.1
https://fastapi.metacpan.org/v1/release/versions/perl?plain=1&versions=5.26.0,5.30.1

### DIFF
--- a/lib/MetaCPAN/Query/Release.pm
+++ b/lib/MetaCPAN/Query/Release.pm
@@ -560,11 +560,28 @@ sub all_by_author {
 }
 
 sub versions {
-    my ( $self, $dist ) = @_;
+    my ( $self, $dist, $versions ) = @_;
 
     my $size = $dist eq 'perl' ? 1000 : 250;
+
+    my $query;
+
+    if ( @{$versions} ) {
+        $query = {
+            bool => {
+                must => [
+                    { term  => { distribution => $dist } },
+                    { terms => { version      => $versions } },
+                ]
+            }
+        };
+    }
+    else {
+        $query = { term => { distribution => $dist } };
+    }
+
     my $body = {
-        query  => { term => { distribution => $dist } },
+        query  => $query,
         size   => $size,
         sort   => [ { date => 'desc' } ],
         fields => [
--- a/lib/MetaCPAN/Server/Controller/Release.pm
+++ b/lib/MetaCPAN/Server/Controller/Release.pm
@@ -86,9 +86,21 @@ sub all_by_author : Path('all_by_author') : Args(1) {
 
 sub versions : Path('versions') : Args(1) {
     my ( $self, $c, $dist ) = @_;
+    my %params = %{ $c->req->params }{qw( plain versions )};
     $c->add_dist_key($dist);
     $c->cdn_max_age('1y');
-    $c->stash_or_detach( $self->model($c)->versions($dist) );
+    my $data = $self->model($c)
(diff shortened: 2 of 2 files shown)


lib/MetaCPAN/Query/Release.pm
```diff
@@ -581,6 +581,16 @@ sub versions {
     my $data = [ map { $_->{fields} } @{ $ret->{hits}{hits} } ];
     single_valued_arrayref_to_scalar($data);
 
+    if ( $params->{versions} ) {
+        my %versions = map { $_ => undef } split /,/, $params->{versions};
+        $data = [ grep { exists $versions{ $_->{version} } } @{$data} ];
+    }
+
+    if ( $params->{plain} ) {
```
haarg, line 589: I think it would make sense to put this part in the controller instead. It's weird to have a function like this return two entirely different forms of data.

oalders (commented):
Thanks for doing this @mickeyn! I think @haarg raises a good point. If there's a simple test we can add to ensure that this doesn't get broken by a subsequent change, that would be helpful.

lib/MetaCPAN/Query/Release.pm
mickeyn: I agree about the plain mode - I'll make that change.

For the versions filtering I think I'll make it a filter in the ES query itself. I don't think that logic shouldn't be in the controller.

haarg: Yeah, the plain mode part is what I meant. The version filtering belongs in the query module.

mickeyn:
I changed the implementation to do the version filtering in the ES query and the plain-mode in the controller.

Still need to add a test, but please re-review.

lib/MetaCPAN/Query/Release.pm
```diff
@@ -560,11 +560,30 @@ sub all_by_author {
...
-    my ( $self, $dist ) = @_;
+    my ( $self, $dist, $versions ) = @_;
 
     my $size = $dist eq 'perl' ? 1000 : 250;
+
+    my $query;
+
+    if ($versions) {
+        $query = {
+            filtered => {
```
haarg, line 571: We should try to use `bool` queries rather than `filtered` queries.  `filtered` has been deprecated in newer ES versions.

mickeyn: You are right, I've been away so long I started to forget stuff:)

Corrected now.

lib/MetaCPAN/Query/Release.pm
```diff
@@ -560,11 +560,28 @@ sub all_by_author {
...
     my $size = $dist eq 'perl' ? 1000 : 250;
+
+    my $query;
+
+    if ($versions) {
+        $query = {
+            bool => {
+                must => [
+                    { term  => { distribution => $dist } },
+                    { terms => { version      => [ split /,/, $versions ] } },
```
haarg, line 574: It might be better to move this string -> arrayref conversion into the controller.

mickeyn: moved

haarg (approved):

oalders (approved):

PR #992, opened 2020-11-16
split examples into paths and files

haarg:
The list of things we're trying to list for examples included 'ex' and
'examples'.  These don't both make sense to match in the same way.

Split the file and path matching. Match directories at the top level,
not just the path prefix. This prevents matching a /ext/ directory. And
match files starting with examples, but not ex.

### DIFF
--- a/lib/MetaCPAN/Query/File.pm
+++ b/lib/MetaCPAN/Query/File.pm
@@ -141,11 +141,19 @@ my %perl_files = (
             )
     ],
 );
-my %prefix_files = (
+
+my %path_files = (
     example => [
         qw(
             eg
             ex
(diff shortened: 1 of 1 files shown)


mohawk2:
@ssoriche The CI here failed with:

```
~/project/metacpan-docker ~/project
ERROR: No such service: api_test
```

@haarg This looks like a great idea. I'm not quite sure about the `path_files` vs `prefix_files` naming? I don't have a better suggestion though. If this PR didn't break the CI, then thumbs up, merge away.

oalders:
We may need to `s/api_test/apitest/` in `.circleci/config.yml`.

Issue #957, opened 2020-05-21
reverse dependencies list incorrectly includes "conflicts"


karenetheridge:
Much to my surprise I found Mouse in Any-Moose's reverse dependencies list. Investigating, it looks like it's coming from this entry in META.json:

https://metacpan.org/source/SKAJI/Mouse-v2.5.10/META.json#L60

"conflicts" is not a normal prerequisite phase and is not indicating something that needs to be or even is suggested as being installed for the distribution. It should not be included in dependency graph calculations.

oalders:
Closed via #996

PR #996, opened 2021-01-16
Fix reverse_dependency query

mickeyn:
Added stricter check for type of relationship.

This should (hopefully) address the issue in #957


mohawk2:
The Travis is failing because `docker: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit.`

haarg:
This query doesn't look correct. Multiple modules will get listed in the `must` clause, but they should be in a `should` because it needs to be a or, not an and.

haarg (approved):

PR #998, opened 2021-01-19
Tidy more files

oalders:
I think there was tidyall caching on my previous PR


mohawk2:
Oh, that would make sense! The caching knows about the subject files being updated, but *doesn't* know about changes to the tidying rules themselves.

mohawk2:
The tidyall is still failing on `t/server/controller/login/pause.t`.

mohawk2 (approved):

PR #1002, opened 2021-02-21
Add a negative boost to Deprecated modules, when searching MetaCPAN.

bleargh45:
Matches modules that are marked as `deprecated: true` in ElasticSearch, and gives them a negative boost in the search results.

Relates to metacpan/metacpan-web#1435

### DIFF
--- a/lib/MetaCPAN/Model/Search.pm
+++ b/lib/MetaCPAN/Model/Search.pm
@@ -222,8 +222,14 @@ sub build_query {
     $params //= {};
     ( my $clean = $search_term ) =~ s/::/ /g;
 
-    my $negative
-        = { term => { 'mime' => { value => 'text/x-script.perl' } } };
+    my $negative = {
+        bool => {
+            should => [
+                { term => { 'mime' => { value => 'text/x-script.perl' } } },
+                { term => { 'deprecated' => { value => 1, boost => -100 } } },
+            ],
+        },
+    };
 
     my $positive = {
         bool => {


lib/MetaCPAN/Model/Search.pm
```diff
@@ -222,8 +222,14 @@ sub build_query {
     $params //= {};
     ( my $clean = $search_term ) =~ s/::/ /g;
 
-    my $negative
-        = { term => { 'mime' => { value => 'text/x-script.perl' } } };
+    my $negative = {
+        bool => {
+            should => [
+                { term => { 'mime' => { value => 'text/x-script.perl' } } },
+                { term => { 'deprecated' => { value => 1, boost => -100 } } },
```
mickeyn, line 229: What is the value 100 based on? for positive matches we boost by 20. 
I'm OK with it, but we do need to see what the effect of it will be and tweak it later if needed.

mickeyn (approved):

lib/MetaCPAN/Model/Search.pm
bleargh45: Honestly, "just a number that I pulled out of thin air".

And yes... **totally** agree that it will be something that we would need to see the effect of on a larger index, and then tweak/adjust as necessary.

oalders: Ok. Let's see what happens. :)

PR #1008, opened 2021-04-04
Author script updates

### LINKED ISSUE #621: Ensure that PAUSE custodial accounts always clobber existing author name
Related to #620  It appears that `AVENJ` is not yet listed with the "custodial" name.  We need to make sure this happens automatically.

mickeyn:
This will 
1. add single-pauseid handling flag to the author script
2. fix name update in case of custodial account

This will fix #621


oalders (approved):
👍

oalders:
I'll fix the tidying in a different PR.

Issue #1009, opened 2021-04-04
handle bus_factor in the river data


neilb:
I've now got my river data generator calculating the **CPAN Bus Factor**, the definition of which was agreed with @oalders.

At the moment, each entry for a distribution is going to look like this:

    {
        "dist" : "Module-Refresh",
        "total" : 25,
        "immediate" : 8,
        "bucket" : 2,
        "bus_factor" : 3
    },

If I'm unable to calculate a bus factor for a distribution, then there won't be a `bus_factor` for that distribution. Is that ok? If you'd rather, I could put `null` in for those cases.

I won't upload new data until you say you're ready for this.

LeoLapworthKT:
But what about the bus factor with CPAN river... is the code in a public repo :) ?

neilb:
> But what about the bus factor with CPAN river

Not sure what this meant?

> is the code in a public repo :) ?

Ha!

mickeyn:
Good news - this is pretty simple.
I only need to update the mapping, so it won't assign `long` by default (will set the field to integer), but once that's done you can add your data where you have the values (no need to include the field where you don't) 
I'll update you soon when mapping is updated and then it's all up to you :)

mickeyn:
And done - you can start including the field.

oalders:
>> But what about the bus factor with CPAN river

> Not sure what this meant?

I believe he's saying the bus number on the CPAN river codebase is currently at 1. ;)

PR #1010, opened 2021-04-04
Added bus_factor to the mapping file

mickeyn:
This is just for future reference, the index (type) mapping is already done.


mickeyn:
related to #1009

oalders (approved):
🚌

PR #1014, opened 2021-04-17
Install Carton along with everything else in snapshot workflow

oalders:


mohawk2 (approved):

mohawk2:
Travis is failing:

```
$ AUTHOR_TESTING=0 cpm install -L $PERL_CARTON_PATH --resolver $CPAN_RESOLVER --workers $(test-jobs) || (tail -n 500 -f ~/.perl-cpm/build.log; false)
FAIL resolve Carton
0 distribution installed.
See /home/travis/.perl-cpm/build.log for details.
You may want to execute cpm with --show-build-log-on-failure,
so that the build.log is automatically dumped on failure.
```

oalders:
I also think there is an issue with the updated dependencies that is triggering a whole bunch of test failures.

PR #1020, opened 2021-05-30
File queries shared by path and prefix for category example

zmughal:
Otherwise directories that are used by releases such as:

  - <https://metacpan.org/source/SRI/Mojolicious-9.18/examples>
  - <https://metacpan.org/source/ETJ/PDL-2.048/Example>

will not have their examples listed on their release page.

Fixes <https://github.com/metacpan/metacpan-web/issues/2427>.

---

Also:

Add demo/demos as category example

For example, found in the release:

  - <https://metacpan.org/source/DCONWAY/Regexp-Grammars-1.057/demo>

### DIFF
--- a/lib/MetaCPAN/Query/File.pm
+++ b/lib/MetaCPAN/Query/File.pm
@@ -142,27 +142,26 @@ my %perl_files = (
     ],
 );
 
+my @shared_path_prefix_examples = qw(
+    example
+    examples
+    Example
+    Examples
+    sample
+    samples
+);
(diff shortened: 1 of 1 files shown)


zmughal:
Another way is to create a hashref `%shared_files` that mirrors the structure of `%path_files`/`%prefix_files` and merge.

haarg (approved):

oalders (approved):
Thanks @zmughal!

PR #1034, opened 2021-07-16
respond to deserialization errors with json

haarg:
If there is an error while deserializing, it is nicer to respond with a
JSON response, rather than plain text. Use a custom Deserialize action
class, which puts the error into the stash where either the Serialize
action class or the JSON view expects. And then skip the rest of the
request.


mohawk2:
Is it possible to make automated tests for this, which otherwise seems like an important improvement? Also, looks like the CI failed because it wants tidying?

oalders (approved):

PR #1038, opened 2021-08-29
fix filters in download_url end point

haarg:
Releases on backpan should not be allowed unless an explicit version is
given. This appears to have been a mistake when converting the query
from cpanminus.


mickeyn (approved):
LGTM

mickeyn (approved):
LGTM

PR #1049, opened 2022-02-22
Capture Exit Code in MetaCPAN::Script::Runner

bodo-hugo-barwich:
To satisfy the requirement for `MetaCPAN::Script::Runner::run()` to return a ` TRUE ` value but also keep track of the Exit Code of the Plugin Modules and probable exceptions that happened in the execution the return value was changed:
```
$ git diff master lib/MetaCPAN/Script/Mapping.pm
diff --git a/lib/MetaCPAN/Script/Mapping.pm b/lib/MetaCPAN/Script/Mapping.pm
index 7e4f47f2..f709aef7 100644
--- a/lib/MetaCPAN/Script/Mapping.pm
+++ b/lib/MetaCPAN/Script/Mapping.pm
@@ -129,7 +129,7 @@ has delete_from_type => (
 );
 
 sub run {
-    my $self   = shift;
+    my $self = shift;
 
     if ( $self->await ) {
         $self->delete_index if $self->arg_delete_index;
@@ -151,9 +151,8 @@ sub run {
         }
     }
 
-    # Correctly reaching this point it will end the application
-    # with the set MetaCPAN::Role::Script::exit_code
-    exit $self->exit_code;
+# The run() method is expected to communicate Success to the superior execution level
+    return ( $self->exit_code == 0 );
 }
``` 
But to communicate the exit code to the main application the `MetaCPAN::Script::Runner::run()` populates a global value `MetaCPAN::Script::Runner::EXIT_CODE` which then can be read by the main application.

To be able to test all this error capturing a _Mock Script_ module was created and a test that checks different error that could happen.

### DIFF
--- a/bin/metacpan
+++ b/bin/metacpan
@@ -6,7 +6,7 @@
 
     bin/metacpan release /path/to/cpan/authors/id/
     bin/metacpan release /path/to/cpan/authors/id/{A,B}
-    bin/metacpan release /path/to/cpan/authors/id/D/DO/DOY/Try-Tiny-0.09.tar.gz  
+    bin/metacpan release /path/to/cpan/authors/id/D/DO/DOY/Try-Tiny-0.09.tar.gz
     bin/metacpan latest
     bin/metacpan server --cpan /path/to/cpan/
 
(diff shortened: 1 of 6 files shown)


bodo-hugo-barwich:
Running the Test Sequence it produce these results:
```

mickeyn (approved):

bodo-hugo-barwich:
@mickeyn  thank you. :)

PR #1051, opened 2022-04-11
Pass the GitHub access token in the Authorization header

voegelas:
GitHub does no longer accept the access token as a query string. The token needs to be passed in the Authorization header instead. Fixes metacpan/metacpan-web#2595 with regard to GitHub.


haarg (approved):

oalders:
Thanks, @voegelas!

PR #1053, opened 2022-05-21
documented_modules should filter for modules with documentation

haarg:
The documented_modules method is only used for one thing currently,
which is finding all of the modules that should be used for link
mappings. This immediately filters for modules that have documentation.
This should instead be part of the filter sent to elasticsearch, to
avoid non-documented files overrunning our result size of 999.


oalders:
This just fails on tidying.

oalders (approved):

Issue #1048, opened 2021-12-18
Functionality to fix a Corrupted ElasticSearch Index


bodo-hugo-barwich:
As documented at
[MetaCPAN API - Indexing failed](https://github.com/metacpan/metacpan-docker/issues/47)
_ElasticSearch_ indices could result corrupted for different reasons.
Often they can be fixed by re-creating the index.
But as seen in the development at
[ElasticSearch Availabilty Check and Mapping Self-Check](https://github.com/metacpan/metacpan-api/pull/1045)
there are still a lot of open questions about how to implement it exactly.

bodo-hugo-barwich:
During the testing it turned out that the **Index Corruption** usually happens when a beginning test is interrupted forcefully:
```
$ docker-compose exec -T api_test prove -lr --jobs 2 t
%hshtestindices{...} in scalar context better written as $hshtestindices{...} at t/lib/MetaCPAN/TestServer.pm line 193.
	require MetaCPAN/TestServer.pm called at t/00_setup.t line 15
	main::BEGIN() called at t/lib/MetaCPAN/TestServer.pm line 193
	eval {...} called at t/lib/MetaCPAN/TestServer.pm line 193
^CERROR: Aborting.
$ docker-compose exec -T api_test prove -lr --jobs 2 t
[Request] ** [http://elasticsearch_test:9200]-[400] [invalid_alias_name_exception] Invalid alias name [cpan], an index exists with the same name as the alias, with: {"index":"cpan_v1_01"}, called from sub MetaCPAN::Script::Mapping::deploy_mapping at /metacpan-api/lib/MetaCPAN/Script/Mapping.pm line 142. With vars: {'status_code' => 400,'request' => {'serialize' => 'std','path' => '/cpan_v1_01/_alias/cpan','ignore' => [],'method' => 'PUT','body' => undef,'qs' => {}},'body' => {'error' => {'reason' => 'Invalid alias name [cpan], an index exists with the same name as the alias','index' => 'cpan_v1_01','root_cause' => [{'type' => 'invalid_alias_name_exception','index' => 'cpan_v1_01','reason' => 'Invalid alias name [cpan], an index exists with the same name as the alias'}],'type' => 'invalid_alias_name_exception'},'status' => 400}}
 at /usr/local/lib/perl5/site_perl/5.30.1/Search/Elasticsearch/Util.pm line 71.
[...]

PR #1045, opened 2021-10-31
ElasticSearch Availabilty Check and Mapping Self-Check

bodo-hugo-barwich:
For the issue detected and documented at
[MetaCPAN API - Indexing failed](https://github.com/metacpan/metacpan-docker/issues/47)
there are several features needed in the `MetaCPAN::Script::Runner` scripts
Seen in this Test Report
[Dependency on _MetaCPAN API_](https://github.com/bodo-hugo-barwich/metacpan-docker/runs/4060036486?check_suite_focus=true#step:5:10)
[_GitHub_ Testing Workflow](https://github.com/bodo-hugo-barwich/metacpan-docker/pull/2)
the successful test depends on a development in _MetaCPAN API_.

This change implements several features:
* in `MetaCPAN::Role::Script`:
  * `exit_code`
  * `cluster_info`
  * `indices_info`
  * `aliases_info`
  * `arg_await_timeout` (triggered with `--await`)
  * `await()` Method
  * `check_health()` Method

The Developer Documentation is also a bit more extended:
```plain
$ perldoc lib/MetaCPAN/Role/Script.pm |more
NAME
    MetaCPAN::Role::Script - Base Role which is used by many command line
    applications

SYNOPSIS
    Roles which should be available to all modules.

OPTIONS
    This Role makes the command line application accept the following
    options

    Option "--await 15"
        This option will set the *ElasticSearch Availability Check Timeout*.
        After "await" seconds the Application will fail with an Exception
        and the Exit Code [112] ("112 - EHOSTDOWN - Host is down") will be
        returned

            bin/metacpan <script_name> --await 15

        See "Method "await()""

METHODS
    This Role provides the following methods

    "await()"
        This method uses the
        "Search::Elasticsearch::Client::2_0::Direct::ping()"
        <https://metacpan.org/pod/Search::Elasticsearch::Client::2_0::Direct
        #ping()> method to verify the service availabilty and wait for
        "arg_await_timeout" seconds. When the service does not become
        available within "arg_await_timeout" seconds it re-throws the
        Exception from the "Search::Elasticsearch::Client" and sets $! to
        112 . The "Search::Elasticsearch::Client" generates a
        "Search::Elasticsearch::Error::NoNodes" Exception. When the service
        is available it will populate the "cluster_info" "HASH" structure
        with the basic information about the cluster. See "Option "--await
        15"" See "Method "check_health()""

    "check_health( [ refresh ] )"
        This method uses the
[...]

PR #1064, opened 2022-07-30
Verify Indices Mappings and Clear all Indices

bodo-hugo-barwich:
This implements the functionality discussed at:
[Fix Corrupted Indices and Verify Mappings](https://github.com/metacpan/metacpan-api/issues/1048)
It is also able to fix difficulties and corruptions as documented at
[Unexpected Index Corruption](https://github.com/metacpan/metacpan-docker/issues/47)

Additionally it implements Exit with Error Code for unconfirmed Operations.

To test all the functionality it features several Mapping Script Tests.

### DIFF
--- a/lib/MetaCPAN/Script/Mapping.pm
+++ b/lib/MetaCPAN/Script/Mapping.pm
@@ -177,7 +177,7 @@ sub run {
         if ( $self->arg_verify_mapping ) {
             $self->check_health;
             unless (
-                $self->verify_mapping(
+                $self->mappings_valid(
                     $self->_build_mapping, $self->_build_aliases
                 )
                 )
@@ -582,7 +582,7 @@ sub deploy_mapping {
     }
 
     $self->check_health(1);
-    $imappingok = $self->verify_mapping( $rmappings, $raliases );
+    $imappingok = $self->mappings_valid( $rmappings, $raliases );
 
     # done
     log_info {"Done."};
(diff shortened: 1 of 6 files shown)


lib/MetaCPAN/Script/Mapping.pm
```diff
@@ -619,12 +983,35 @@ with B<Exit Code> C< 1 >.
...
-See L<Method C<verify_mapping()>>
+See L<Method C<check_mapping()>>
 
 See L<Method C<MetaCPAN::Role::Script::check_health()>>
 
 =back
 
+=cut
+
+#<<<<<<< HEAD
```
haarg, line 994: Some left over conflict markers.

haarg (changes requested):

lib/MetaCPAN/Script/Mapping.pm
bodo-hugo-barwich: Thank you for your interest in my work.

Yes, you are right. 
I looked into it and fixed the block of documentation which was meant to explain the `--delete --all` option.

```
    Option "--all"
        This option is only effective in combination with Option "--delete".
        It uses the information gathered by
        "MetaCPAN::Role::Script::check_health()" to delete ALL indices in
        the *ElasticSearch* Cluster. This option is usefull to reconstruct a
        broken *ElasticSearch* Cluster

            bin/metacpan mapping --delete --all

        See "Option "--delete""

        See "Method "deploy_mapping()""

        See "Method "MetaCPAN::Role::Script::check_health()""
```

bodo-hugo-barwich: As documented in the related issue at
[Fix Corrupted Indices](https://github.com/metacpan/metacpan-api/issues/1048)
this will only work if the **Environment Variables** `PLACK_ENV` or `MOJO_MODE` are set to `development` or `testing`.
Otherwise this will throw an **Exception** `[ 1 ] - EPERM - Operation not permitted`.

mohawk2:
This is a big (adding 700 lines) change that introduces valuable functionality. My only observation is it seems splittable conceptually into two pieces of functionality; the verify, and the clear-indices. Would it make sense to make a separate PR that only introduced the checking/verification code from this? That would be a smaller, much less risky change; then this one could be rebased past that, with the higher-risk code isolated and more easily reviewable.

bodo-hugo-barwich:
Yes, you are right. I also noticed that I actually was working on two different tasks. But unfortunately both share the same code and also the same tests. So I need to rebuild a separate branch from the master branch and reimport the sub routines from the current branch, and review it of course. Also the second branch would need to be rebased according to which branch is merged first.

bodo-hugo-barwich:
I built a new branch at:
[Clear Unknown Indices](https://github.com/metacpan/metacpan-api/pull/1065)
which implements the `--delete --all` option and also introduces important tests which are used for the mappings verification tests.

mohawk2:
I would really like to see the verify done (including merged and released, since it is inherently less risky) first, not the clear. Is there a good reason why it can't be done in that order?

bodo-hugo-barwich:
Actually the verify mappings logic is the bigger part of this development with the introducing the `_compare_mapping()`, `_build_mapping()` and `_build_aliases()` methods with more than 200 lines of code additionally to the tests with another 100 lines of code.
The tests of the verify mappings logic require tests of `--create` and `--update` index functionality which require some additional 100 lines of changes.
So my idea was to **split the code more evenly** and already introduce tests in the smaller task that will **thin out** the verify mappings development which supposedly would increase the maintainability and reviewabilty of the development.

On the other side the tests of the `--delete --all` logic show that this functionality is **save and should not be able to run in production** if my assumptions about the environment recognition are as documented in the related issue:
[recognize development environment](https://github.com/metacpan/metacpan-api/issues/1048#issuecomment-1186163526)
[flag development environment with environment variables](https://github.com/metacpan/metacpan-api/pull/1045#discussion_r767159095)
Still there is also a manual protection as requested at:
[manual confirmation required](https://github.com/metacpan/metacpan-api/pull/1045#discussion_r749563851)

bodo-hugo-barwich:
I rebuilt the "_Verify Mappings_" functionality at:
[Verify Indices Mappings](https://github.com/metacpan/metacpan-api/pull/1067)
which is the semantical split of this development.

mickeyn:
the other PR is merged, so shall we close this one?

bodo-hugo-barwich:
this development was split into 2 separate Pull Requests

 * [Clear Unknown Indices](https://github.com/metacpan/metacpan-api/pull/1065)
 * [Verify Indices Mappings](https://github.com/metacpan/metacpan-api/pull/1067)

 so this Pull Request is obsolete and can be closed

PR #1081, opened 2023-01-03
fix reverse dependency query

haarg:
The reverse dependency query attempted to find releases with a dependency of a given name and a relationship of 'runtime'. But because the dependencies are stored as an array and it was not doing a nested query, it was actually searching for a release with a dependency of a given name, and any other dependency with a relationship of 'runtime'.

Fix this by using a nested query for the dependency array, and include an additional filter on the phase. We don't want to include reverse dependencies of "develop" or other custom phases.


mickeyn (approved):
LGTM

PR #1091, opened 2023-04-28
update snapshot updating action to match metacpan-web

haarg:
This updates the GitHub actions to better match the behavior of metacpan-web. This includes changing the comment used on the PR, updating the formatting, and enabling automerge.


oalders:
I think this will get the build past the error:

```
diff --git a/Dockerfile b/Dockerfile
index 138e7d90..c34e13cf 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@ WORKDIR /metacpan-api
 # size of the images.
 RUN mkdir /CPAN \
     && apt-get update \
-    && apt-get install -y --no-install-recommends rsync=3.1.3-6 jq \
+    && apt-get install -y --no-install-recommends rsync jq \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/* \
     && cpm install --global \
```

oalders (approved):

PR #1105, opened 2023-05-03
fix numeric sort

mickeyn:
this fixes the warnings: "Argument "1.38_03" isn't numeric in numeric comparison (<=>)"

### DIFF
--- a/lib/MetaCPAN/Script/CVE.pm
+++ b/lib/MetaCPAN/Script/CVE.pm
@@ -179,11 +179,12 @@ sub index_cve_data {
 
                 if ( $releases->{hits}{total} ) {
                     ## no critic (ControlStructures::ProhibitMutatingListFunctions)
-                    @matches = sort { $a->{version} <=> $b->{version} }
+                    @matches = map { $_->[0] }
+                        sort { $a->[1] <=> $b->[1] }
                         map {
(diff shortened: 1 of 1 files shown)


lib/MetaCPAN/Script/CVE.pm
```diff
@@ -179,7 +179,10 @@ sub index_cve_data {
 
                 if ( $releases->{hits}{total} ) {
                     ## no critic (ControlStructures::ProhibitMutatingListFunctions)
-                    @matches = sort { $a->{version} <=> $b->{version} }
+                    @matches = sort {
+                        numify_version( $a->{version} )
+                            <=> numify_version( $b->{version} )
+                        }
```
atoomic, line 185: maybe consider a local cache for `numify_version` or not 🤷

mickeyn: sure, but I didn't think the runtime saved would be worth the extra memory used. this script takes <1m. :)

mickeyn: @atoomic please check the newer version

atoomic (approved):

haarg:
This is fine, but wasn't the version already supposed to be numified?

mickeyn:
@haarg version field is the raw info, version_numified is the one we process

Issue #1129, opened 2023-09-03
`/release/by_author` is missing a `from` or `page` option


jackdeguest:
The endpoint `/release/by_author`[1] usually returns a lot of releases, and the option `size` allows defining the `size` of each data set returned, but there is no option to navigate among those data set, such as an option `from` or `page`.

For example [/release/by_author/OALDERS](https://explorer.metacpan.org/?url=/release/by_author/OALDERS) returns 556 hits.

[1] [https://github.com/metacpan/metacpan-api/blob/88260ad8cc71aeff93fe673bdf7322a9700b61b1/lib/MetaCPAN/Server/Controller/Release.pm#L62](https://github.com/metacpan/metacpan-api/blob/88260ad8cc71aeff93fe673bdf7322a9700b61b1/lib/MetaCPAN/Server/Controller/Release.pm#L62)

mickeyn:
#1131 will solve this.

Future work - merge the queries from multiple subs in the files & sub arguments naming standardization (author, pauseid, etc.)

jackdeguest:
> Future work - merge the queries from multiple subs in the files & sub arguments naming standardization (author, pauseid, etc.)

An excellent idea. For example, the following endpoints, use the option `page_size` where it is usually just `size` elsewhere.

* `/v1/release/all_by_author/{author}`
* `/v1/release/recent`
* `/v1/release/top_uploaders`
* `/v1/reverse_dependencies/dist/{distribution}`
* `/v1/reverse_dependencies/module/{module}`

PR #1131, opened 2023-09-03
release/by_author: support 'page' parameter

mickeyn:
Adding a page parameter support for /release/by_author
solves #1129


oalders (approved):
🚀

PR #1165, opened 2024-02-20
log CDN purges

haarg:
The number of purges at Fastly seems very large. Logging the purges should help track down why.


oalders (dismissed):

oalders:
> Global symbol "$c" requires explicit package name (did you forget to declare "my $c"?) at /metacpan-api/lib/MetaCPAN/Server.pm line 156.
> Global symbol "@keys" requires explicit package name (did you forget to declare "my @keys"?) at /metacpan-api/lib/MetaCPAN/Server.pm line 159.

oalders (approved):

PR #1180, opened 2024-04-08
author import: handle duplicate ids

haarg:
When reading 00whois.xml, there can be cpanid elements containing duplicate id element values. This can happen when there is both a list and author entry for the same name. In these cases, the author entry is extraneous.

Rather than relying on XML::Simple's built in array to hash transformation (using id values), do the work ourselves so we can handle the duplicate id entries. Also configure XML::Simple to be more strict about how it parses.

### DIFF
--- a/lib/MetaCPAN/Script/Author.pm
+++ b/lib/MetaCPAN/Script/Author.pm
@@ -78,9 +78,49 @@ my @compare_fields = do {
     sort grep !$seen{$_}++, @cpan_fields, @author_config_fields;
 };
 
+has whois_data => (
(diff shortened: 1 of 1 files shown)


mickeyn (changes requested):
I don't see any records with `<type>list</type>` owning anything in `01modules` - maybe the solution should just be sticking to records with `<type>author</type>` ?

haarg:
Mailing list users are not able to upload files, but they are able to hold permissions.

mickeyn (approved):

PR #1186, opened 2024-04-25
fix download_url end point querying for releases

haarg:


lib/MetaCPAN/Query/Release.pm
```diff
@@ -1059,8 +1059,10 @@ sub find_download_url {
             must => [
                 { term => { $prefix . 'authorized' => 1 } },
                 { term => { $prefix . 'indexed'    => 1 } },
-                { term => { $prefix . 'name'       => $module } },
                 (
+                    $module_filter ? { term => { $prefix . 'name' => $name } }
+                    : { term => { 'release' => $name } },
+                ) (
```
oalders, line 1065:
```suggestion
                ),
```

oalders: Haven't tested, but this line seems to be causing the test failures.

oalders (changes requested):

oalders (approved):

PR #1187, opened 2024-04-26
actually really fix download_url end point for perl releases

haarg:


lib/MetaCPAN/Query/Release.pm
```diff
@@ -1176,10 +1189,17 @@ sub find_download_url {
...
-    };
+    my $source_name = delete $source->{name};
+    if ( !$module_filter ) {
+        $source->{release} = $source_name;
+    }
+
+    my $module
+        = $hit->{inner_hits}{module}
+        ? $hit->{inner_hits}{module}{hits}{hits}[0]{_source}
+        : {};
```
atoomic, line 1200: probably always want to add a `$module //= {}` as we are going to dereference it later

lib/MetaCPAN/Query/Release.pm
```diff
@@ -1176,10 +1189,17 @@ sub find_download_url {
...
+
+    return +{ %$source, %$module, @checksums, };
```
atoomic, line 1202: note that `$source` can be `undef`

haarg: At this point, we have verified that there was a result in `$hit`. We are requesting specific `_source` fields. Even if none of those fields are available, `_source` will still exist unless Elasticsearch is entirely broken. `$source` can not be `undef`.

lib/MetaCPAN/Query/Release.pm
haarg: The ES query that would produce `{inner_hits}{module}` requires a module with a matching name. If `inner_hits` exists, `{inner_hits}{module}{hits}{hits}[0]{_source}` will always exist.

oalders (approved):

PR #1202, opened 2024-04-28
Move Elasticsearch server setup to config files

oalders:
- **Make config easier to get for now**
- **Get elasticsearch servers from config**


mickeyn (dismissed):

lib/MetaCPAN/Script/Runner.pm
```diff
@@ -3,9 +3,10 @@ package MetaCPAN::Script::Runner;
 use strict;
 use warnings;
 
-use Config::ZOMG        ();
-use File::Path          ();
-use Hash::Merge::Simple qw( merge );
+use File::Path               ();
+use MetaCPAN::Server::Config ();
+use Module::Pluggable        qw( plugins );
```
haarg, line 8: Here and in several other places duplicate use statements have been added.

t/script/cover.t
```diff
@@ -1,14 +1,18 @@
 use strict;
 use warnings;
+
 use lib 't/lib';
 
 use MetaCPAN::Script::Cover  ();
-use MetaCPAN::Script::Runner ();
+use MetaCPAN::Server::Config ();
 use MetaCPAN::Util           qw( checkout_root );
+use Test::More import => [qw( done_testing ok )];
```
haarg, line 9: This is a duplicate use, but also tries to enumerate the imports to use. That's a good idea in general, but I don't like it for Test::More, and we have it excluded from the import checks.

oalders: Yeah, not sure how this one happened.

mickeyn (approved):

PR #1205, opened 2024-04-30
drop support for metacpan_script

haarg:
We no longer need metacpan_script for anything our code uses, and supporting it complicates our query filtering and interactions with Elasticsearch.

We've generally been trying to move away from users interacting via Elasticsearch queries, but if they really need to and need a specific script, they are always welcome to submit a PR to the API adding what they need.


oalders (approved):

oalders (approved):
I assume if there's some external thing that relies on this we could temporarily back out of the change. I think we need to remove the JOIN documentation as well.

oalders:
Wrong PR for that comment ^^.

PR #1210, opened 2024-05-06
remove use of Pod::POM and instead use Pod::Simple::JustPod

haarg:


mickeyn (dismissed):

oalders:
Looks like an issue with line endings in the diff in `t/pod/renderer.t `

PR #1230, opened 2024-06-19
Clean up many prereqs

haarg:
Remove a bunch of unused prereqs.

Also replace some prereqs with simpler variants.


lib/MetaCPAN/Role/HasConfig.pm
```diff
@@ -24,13 +25,12 @@ sub _build_config {
     my $config = $self->_zomg("$FindBin::RealBin/..");
     return $config if $config;
 
-    require_module('Git::Helpers');
-    $config = $self->_zomg( Git::Helpers::checkout_root() );
+    $config = $self->_zomg( checkout_root() );
```
oalders, line 28: This removes the `require` on `Git::Helpers`, but still uses the `checkout_root()` function, so we'll need to add it to the `use` statements above.

oalders: Oh, I see it has moved to `MetaCPAN::Util`.

oalders:
Looks like it wants `Data::Printer` in the tests.

mickeyn (dismissed):

PR #1245, opened 2024-09-02
A large CircleCI VM can able to handle 4 test jobs

oalders:
This PR:
* reduces the CircleCI build time by ~10%
* simplifies the CircleCI config by using an official Orb rather than deprecated code


ranguard (approved):
Find having `name` _after_ what it's doing a bit odd.. I want to read what it's mean to be doing and _then_ how it's doing it... but that's how it was before

PR #1256, opened 2024-09-20
Switch configs to YAML

oalders:


oalders:
I have diffed the Apache-style configs with these new YAML configs and they are consistent with each other.

mickeyn (approved):

PR #1261, opened 2024-09-24
remove use of search_type scan

haarg:
search_type: scan has been removed from newer Elasticsearch versions,
replaced with sort: _doc.


haarg:
Fixes #1262

oalders (approved):

PR #1282, opened 2024-10-02
remove index option from scripts and reimplement it for backup

haarg:
The index option is meant to control the index that has all of our ES data. But in the future, each document type will need to be in a separate index. So the idea of having a global configurable index doesn't make sense.

The backup script needs to be able to operate on all indexes, and works generically with any type. Reimplement an index option for it, allowing multiple indexes to be specified.

### DIFF
--- a/bin/cron/author.sh
+++ b/bin/cron/author.sh
@@ -3,7 +3,4 @@
 # export ES_SCRIPT_INDEX=author_01
 # /home/metacpan/bin/metacpan-api-carton-exec bin/metacpan author --index author_01
 
-export ES_SCRIPT_INDEX=cpan_v1_01
-/home/metacpan/bin/metacpan-api-carton-exec bin/metacpan author --index cpan_v1_01
-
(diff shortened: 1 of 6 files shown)


haarg:
With this PR, the `backup` script only requires the `--index` option, instead of also requiring the `ES_SCRIPT_INDEX` environment variable. None of the other scripts have any use for the `--index` option or the `ES_SCRIPT_INDEX` env var, because they are either using `cpan_v1_01` or their own index that is defined in the code.

haarg:
I guess I didn't entirely answer your question. No, this isn't only assuming we only use the `cpan` index. It's just making it so the use of `cpan_v1_01` is not configurable. This is so the rest of the code can be updated to refer directly to their index, rather than trying to use a dynamic value. I'm intending to set up some infrastructure to make it easy to switch the index/type of each of the current "types", so we can move to one type per index.

mickeyn (approved):

PR #1286, opened 2024-10-08
Don't assume an author is male

oalders:


jackdeguest:
Hey guys, I am no English experts, but since you are perfecting your already great piece of software, shouldn't this: `# The author can manage this identity both on their gravatar account` rather be `# The authors can manage this identity both on their gravatar account`.

PR #1290, opened 2024-10-15
update CVE data urls

haarg:
The old URLs have stopped working.

I'm not certain that the new URLs are the correct ones, but they seem to be getting updated regularly.

@stigtsp


mickeyn (approved):

stigtsp:
The new URL is the correct one, thx!

PR #1292, opened 2024-10-21
various cleanups

haarg:
Remove some unused prereqs.

Simplify the /v1/ URL handling in dev mode by using a mount, rather than a rewrite.

Move the search module to a query module, since it essentially matches the pattern of those modules.

Calculate the root directory based on the files, not based on git.


oalders:
Is this ready for review?

oalders (approved):
⭐

PR #1334, opened 2024-12-28
Move more queries to Query namespace, with parent object to call between modules

haarg:
Adds a parent object to hold all of the query objects, allowing them to easily access each other. Use this object in most of the controller classes, rather than using the MetaCPAN::Model instance.

Remove all of the no longer used delegated methods in the model set classes. For the set classes that no longer have any of their own methods, remove them.

Move several methods from Model to Query.


mickeyn (approved):
LGTM!

Issue #165, opened 2012-02-18
Games::Poker::TexasHold::em does not appear to be indexed


grantm:
No matches: https://metacpan.org/search?q=Games%3A%3APoker%3A%3ATexasHold%3A%3Aem

Tarball is on CPAN and indexed on s.c.o: http://search.cpan.org/dist/Games-Poker-TexasHold-em/

PR #1344, opened 2025-02-01
normalize package names to use ::

haarg:
package names extracted by PAUSE or Parse::PMFile are normalized to use double colons rather than apostrophe. Module::Metadata does not. Add the normalization to the output of Module::Metadata when we index files so that the module names will match what exists in 02packages.


haarg:
This should fix #165.

mickeyn (approved):

PR #1363, opened 2025-03-12
precious tidy

oalders:


mickeyn (approved):

lib/MetaCPAN/Util.pm
```diff
@@ -18,7 +18,6 @@ use Ref::Util      qw(
     is_plain_hashref
     is_ref
 );
-use Cpanel::JSON::XS ();
```
haarg, line 21: This removal is wrong. There are subs manually imported from this package later in the file.

haarg (changes requested):

mickeyn (approved):

PR #1377, opened 2025-06-14
Update to latest MC::Role so we get stale-while-revalidate header

ranguard:


oalders:
I am running the snapshot updater on this branch, which should reduce the number of lines changed, but I see that action has been failing for a while due to `Search::Elasticsearch`, so we may need to figure that out before deploy this branch.

oalders:
@haarg any opinion on what we should pin to?

PR #1383, opened 2025-06-28
Docker build and other cleanups

haarg:
Update the Dockerfile and build process for better caching and to prepare to deploy in Kubernetes.

A docker compose configuration was added to make local testing easier.

Clean up various old files that are no longer relevant.

Switch to precious for formatting and remove formatting test. Replace with a GitHub workflow.

Remove Mojolicious web code. Only Minion and the admin interface remains. Most of the end points were never ported, and MountPSGI ended up being problematic in a number of ways. Production has been switched back to Gazelle for a while now, just using app.psgi. For our other systems deployed in k8s, we're using uwsgi and should be standardizing on it. If someone wants to do a proper port to Mojolicious that could be considered, but it needs to be more complete and would probably only make sense once we're fully disconnected from the ingestion side. We're still using Minion, but it shouldn't be part of the API server.

The CircleCI config has been revised to use the docker compose configuration from this repo rather than pulling it from another. This should make things simpler to update in the future.


haarg:
With this merged, tests can be run locally by checking out the repo and running:
```sh
docker compose --profile test run --build api-test prove -lrv t
```

ranguard (approved):

oalders (approved):
I had a quick look and I don't have any concerns. Are the coverage reports still working?

Given where we are at, I'm ok with "move fast and break things" as far as local dev goes. I will try to mess around with the dev env as we merge things to see if I spot any issues.

haarg:
Coverage should still be working fine

oalders:
> Coverage should still be working fine

I don't see anything at https://app.codecov.io/gh/metacpan/metacpan-api/tree/haarg%2Fcleanups but that's not so important right now.

PR #1426, opened 2026-02-24
Update purge script for Path::Tiny

oalders:
- **the cpan method in Purge.pm now provides a Path::Tiny object**
- **clarify that the purge script needs an author when deleting a release**


oalders:
That failure is just saying that the coverage has not improved.

PR #1429, opened 2026-02-27
refactor types to all use Type::Tiny

haarg:
Refactor type definitions to all use Type::Tiny. Move the document specific types into their own modules. The general types go into MetaCPAN::Types::Internal. And MetaCPAN::Types merges all of those with a number of type libraries from CPAN for easy use.

Several of the Elasticsearch objects had sub-objects which were just glorified hashes. These have been removed.

The Module sub-object has some behavior attached to it, so it has been maintained.

The coercion from a single item to an arrayref is now done with a parameterized coercion. This is a bit of a hack, but easily allows the coercion to be applied to an arrayref of other types, and coerce to those properly.


lib/MetaCPAN/Role/Logger.pm
```diff
@@ -5,9 +5,8 @@ use Moose::Role;
 
 use Log::Contextual qw( set_logger );
 use Log::Log4perl ':easy';
-use MetaCPAN::Types::TypeTiny qw( Logger Str );
```
oalders: Does this mean that `--level` can no longer be provided to scripts?

oalders (approved):
🧹

lib/MetaCPAN/Role/Logger.pm
haarg: I don't see how this PR changes anything related to that.

The level option is provided by this role because it is composed by MetaCPAN::Role::Script.

PR #1450, opened 2026-03-23
use libarchive for accessing source of files in releases

haarg:
libarchive can extract any of the formats we need, is more performant and uses less memory than the perl implementation used by Archive::Any. It can also properly protect against symlinks or absolute paths in tar files.

Archive::Any is still used by the indexing code. That will be replaced in the future.


oalders (approved):
Is this something that ingest will need as well?

PR #1470, opened 2026-04-23
Test setup and other various cleanups

haarg:


lib/MetaCPAN/Model/Email/PAUSE.pm
```diff
@@ -26,12 +28,14 @@ has _url => (
 );
 
 sub send {
-    my $self = shift;
+    my $self  = shift;
+    my $email = $self->_author->{email};
```
oalders: This addition means that the variable gets shadowed just below, which seems a bit confusing.

lib/MetaCPAN/Script/Role/Contributor.pm
```diff
@@ -85,18 +85,30 @@ sub release_contributor_update_actions {
...
-    my @docs     = map {
-        ;
-        my $contrib = $_;
-        {
+
+    for my $contrib (@$contribs) {
+        my $id = digest(
+            $release->{name}, $release->{author}, $contrib->{pauseid},
+            $contrib->{name}, $contrib->{email},
+        );
```
oalders, line 94: Is the value of `$id` deterministic?

oalders:
I think the codecov failure may just be a transient thing.

lib/MetaCPAN/Model/Email/PAUSE.pm
haarg: Fixed.

lib/MetaCPAN/Script/Role/Contributor.pm
haarg: digest is doing a SHA-1 hash, so this should make it deterministic.

oalders: What if `$contrib->[email]` is an arrayref?

PR #1491, opened 2026-06-14
Fetch GitHub stars for any GitHub-hosted dist (and fix index_github_bugs no-op)

### LINKED ISSUE #1477: github star counts only fetched if github issues used
The GitHub star count is only fetched and stored for dists that list GitHub issues as their bug tracker. If the repository is listed as using GitHub but the issue tracker is RT, the star count is never fetched.

oalders:
Closes #1477

## Problem

GitHub star/watcher counts were only fetched for distributions whose **bug tracker** was GitHub issues. A dist that uses RT (or any non-GitHub tracker) but hosts its code on GitHub never had its star count fetched.

While fixing this I found a deeper bug: `index_github_bugs` has been a **silent no-op since November 2024**. Commit 7c471c01 migrated the loop from a model scroll (`$release->resources`) to a raw Elasticsearch `scroll_helper` but kept top-level hash access (`$release->{resources}`, `$release->{distribution}`). Raw ES hits store fields under `_source`, so both reads returned `undef`, every release was skipped, and **no GitHub bug or star data was being updated at all**.

## Changes

- **Read release fields from `_source`** in `index_github_bugs`, matching `check_all_distributions` and every other script. This restores the feature.
- **Decouple star fetching from bug counting**: the release query now also matches `resources.repository.url` / `resources.repository.web`, and `_github_dist_summary` records stars/watchers for any dist with a GitHub repo while recording GitHub issue counts only when GitHub issues are the bug tracker.
- **`_is_github_url`** parses the URL and compares its host to `github.com` (via `URI->new->host`) instead of a prefix match, rejecting look-alike hosts such as `github.com.evil.com`. Adds `URI::git` (cpanfile + snapshot) so `git://` URLs resolve a host.
- The `git://` scheme is matched only against repository URL fields (bug trackers are always http(s)).
- Skip releases lacking a `distribution`, and skip a null `repository` in an otherwise-successful GraphQL response before dereferencing it.

## Testing

- New `t/script/tickets.t` unit tests for the host check, the query filter, and the star-vs-issue decision logic (including the RT-bugtracker-plus-GitHub-repo case that motivates the fix).
- `prove t/script/tickets.t t/script/load.t` passes; `precious lint` clean (perlimports/perlcritic/perltidy).
- Red-green verified: simulating the old coupled behavior makes the decoupling tests fail.
- `cpm install --resolver=snapshot URI::git` resolves from the new snapshot stanza.
- Not verified end-to-end: a live `index_github_bugs` run needs a GitHub token and ES release fixtures, which weren't available here. The `_source` fix is verified by consistency with the rest of the codebase rather than an integration test.

### DIFF
--- a/lib/MetaCPAN/Script/Tickets.pm
+++ b/lib/MetaCPAN/Script/Tickets.pm
@@ -12,6 +12,7 @@ use Net::GitHub::V4       ();
 use Ref::Util             qw( is_hashref is_ref );
 use Text::CSV_XS          ();
 use URI::Escape           qw( uri_escape );
+use URI::Split            qw( uri_split );
 
 with 'MetaCPAN::Role::Script', 'MooseX::Getopt';
 
@@ -186,7 +187,18 @@ END_QUERY
         }
 
         my $repo_data = $data->{data}{repository};
-        my $github    = $self->_github_dist_summary( $resources, $repo_data );
+
+        # A successful response can still carry a null repository (e.g. a repo
+        # that became private without an explicit NOT_FOUND error). Skip it
+        # rather than dereferencing undef.
+        unless ($repo_data) {
+            log_info {
+                "[$release->{distribution}] no repository data returned"
+            };
+            next RELEASE;
+        }
+
+        my $github = $self->_github_dist_summary( $resources, $repo_data );
 
         $dist_summary->{'repo'}{'github'} = $github->{repo}{github};
         $dist_summary->{'bugs'}{'github'} = $github->{bugs}{github}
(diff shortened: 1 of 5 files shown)


lib/MetaCPAN/Script/Tickets.pm
```diff
@@ -196,51 +191,152 @@ END_QUERY
...
-            if is_hashref($v);
+    return $self->_github_search_resources($resources);
+}
 
-        return ( $user, $repo, $source ) if $user;
+# Recursively walk the resources hash for any github url.
+sub _github_search_resources {
+    my ( $self, $resources ) = @_;
+
+    for my $v ( values %{$resources} ) {
```
haarg, line 331: Would be better to use sorted keys so this is consistent.

mickeyn (approved):

PR #1500, opened 2026-09-15
Keep all valid author emails

oalders:
author.json and profile edits can supply email as an arrayref, which stringified
to ARRAY(0x..), failed Email::Valid, and got reset to <pauseid>@cpan.org. Unwrap
it like website and keep every valid address; fall back only when none validate,
now with a warning.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>


oalders:
This is only a partial fix, but I thought it would be easier to break this up. Obviously the @cpan.org fallback is misleading at best, but we can tackle that in a different PR.

mickeyn (approved):

PR #1502, opened 2026-09-16
use hash for permission check rather than grepping array

haarg:
All uses of the permissions structure end up searching over its entries. A hash is a better structure for this use case.


lib/MetaCPAN/Document/File.pm
```diff
@@ -985,15 +985,13 @@ sub set_authorized {
         foreach my $module ( @{ $self->module } ) {
             $module->_set_authorized(false)
                 if ( $perms->{ $module->name }
-                && !grep { $_ eq $self->author }
-                @{ $perms->{ $module->name } } );
+                && !$perms->{$module->name}{$self->author};
```
oalders: syntax error at /app/lib/MetaCPAN/Document/File.pm line 988, near "};"

oalders (approved):

PR #1503, opened 2026-09-17
Stop author import from overwriting UI profile edits

oalders:
## Problem

The hourly author import (`bin/metacpan author`) rebuilds each author document from PAUSE whois + `author-*.json` and reverts any edits an author made through metacpan.org.

## Fix

When an author saves their profile, `profile_PUT` stamps `canonical_source => 'ui'` on the document. The import reads that flag and skips those authors, so UI edits stick.

- `canonical_source` is a `source_only` Str field — no ES mapping migration needed (`dynamic: false`).
- The controller is the only writer; the cron only reads.
- Kept out of `@compare_fields` (a control flag, not diffed content).

## Custodial accounts

There's one exception to "leave UI-canonical authors alone": when PAUSE hands an account to a custodian it rewrites the whois `fullname` (appending `(PAUSE Custodial Account)`). That name is how people learn the account was handed off, so it has to reach **every** author, including UI-canonical ones the import would otherwise skip.

- For a UI-canonical author, a custodial rename writes **only** the name — the rest of their profile is left exactly as they curated it.
- For everyone else, the custodial whois name wins over any name from `author.json`.
- Nothing else changes (email and other contact fields are untouched — we are not erasing anyone), and the flag is not flipped.
- Self-stabilizing: once the stored name matches the custodial name, no further writes are issued.

## Not otherwise reset (intentional)

Outside the custodial case, the `ui` flag is permanent — the import will not apply later PAUSE/`author.json` changes to a UI-canonical author. That's intended: the UI is canonical. A future UI control could clear the flag to resume automatic updates; the cron guard already falls through for any value other than `'ui'`, so no cron change would be required.

## Testing

`t/script/author-canonical-source.t` covers the guard and the custodial path (ui doc untouched, new author rebuilt, existing author without the flag rebuilt, ui-canonical custodial gets a name-only update, an already-renamed custodial account is left alone, and the custodial name beats `author.json`) via a fake bulk helper — no Elasticsearch required.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Grinnz:
Would this result in missing any important updates from upstream that aren't editable in metacpan?

oalders:
From PAUSE whois we get:

- fullname
- asciiname
- email
- homepage

From `author.json` we can get arbitrary data. There is a chance that something doesn't get merged in, but the current situation is worse for people who use the MetaCPAN UI to update their profiles. The profile updates successfully and then gets silently clobbered on the next cron run. 

Dealing with 3 different sources, trying to give them priority and then trying to figure out which is most up to date so that it can be re-prioritized is not going to be easy and so far we haven't really gotten it right.

The first problem I want to solve is letting authors update their info via the UI and ensuring that gets preserved. Right now if your email gets set back to your @cpan.org address it gets that much harder to contact you about a CVE. Not impossible, but harder than it needs to be.

We probably should consider an account that flips to PAUSE custodial and remove the UI as the canonical source in that case. Would need to think about that a bit more.

Grinnz:
Would it be possible to store both separately, by source rather than one merged dataset? Not sure if it would be reasonable to show both either, just thinking of ways to preserve having a "canonical" external data set without having it end up getting discarded

oalders:
I think so, but I was trying to find the simplest way to move forward for now. The whois and `author.json` are actual text records, so the only thing really getting discarded right now is the UI form submission. I think we could look at something like that in another iteration, but it really is not great to have to juggle the 3 sources and I'm guessing in most cases we can just rely on one or two of them.

ranguard (approved):
Seems reasonable!

